PR #5, opened 2021-12-07
Scotty bring u tp into monorepo

ScottyPoi:


acolytec3:
It looks like you're missing a ``UtpAlgConfiguration` file somewhere.  There are two places (`src/wire/utp/Itls/constants.ts` and `src/wire/utp/Protocol/congestionControl/StatisticsLogger.ts`) that import it but I don't see it anywhere in the code.

packages/portalnetwork/src/wire/utp/Protocol/read/UtpReadListener.ts
```diff
@@ -0,0 +1,34 @@
+// import { Worker,  } from 'worker_threads';
+import cluster, {Worker} from 'cluster'
```
acolytec3, line 2: Were you able to get this to compile and run in browser?  This is a nodejs module and looks like it's not available in the browser, at least according to [this PR](https://github.com/facebook/create-react-app/pull/6010)

PR #11, opened 2021-12-11
A FIND CONTENT request will run test data

ScottyPoi:
As of 12/10 --

Test will run, but not yet successfully decoding the payloads on receivers side.

The correct number of bytes will be sent, over the correct number of packets, yet the payloads decode to arrays of zeros for unknown reasons.

Nodes will do FIN exchange and stop sending packets.  No actual "closing" of utp sockets happens.

For test purposes, we're sending a Uint8Array of 2000 1's.  The size limit is set to 100 bytes per packet.  In real practice, this will be between 150 and 1200


acolytec3 (approved):
It works (at least for happy path!)

PR #15, opened 2021-12-13
Replace discv5 packge with subtree from upstream

acolytec3:
Replaces our ported/renames `portalnetwork-discv5` with the my original fork of `discv5` now with a `git subtree` implementation that will allow us to merge commits from the upstream `chainsafe/discv5` library.


acolytec3:
@ryanio @ScottyPoi Can one or both of you try cloning this branch and building and see if everything installs and builds correctly?  I found a useful guide [here](https://www.atlassian.com/git/tutorials/git-subtree) on `git subtree` which should allow us to maintain the discv5 linked to Chainsafe's upstream version and pull in new commits from that module as needed.

acolytec3:
Going to optimistically merge this based on confirmation from @ScottyPoi that it doesn't appear to break anything and I was able to add a commit to my upstream fork as designed.

PR #21, opened 2021-12-16
Succesfully sends / receives and recompiles  a test vector

ScottyPoi:
Also cleaned up the directory, attempting to remove unused code.
Added comments and debug logs to help illuminate the process.


packages/portalnetwork/src/client/client.ts
```diff
@@ -368,18 +374,10 @@ export class PortalNetwork extends (EventEmitter as { new(): PortalNetworkEventE
 
     // private handleContent = async (srcId: string, message: Italk)
 
-    private handleUTPStreamRequest = async (srcId: string, msgId: bigint, packetBuffer: Buffer) => {
-
+    private handleUTP = async (srcId: string, msgId: bigint, packetBuffer: Buffer) => {
+        await this.client.sendTalkResp(srcId, msgId, new Uint8Array())
```
acolytec3, line 378: What is this line doing?  As far as I know, the current state of the uTP spec is to not ever use the `TALKRESP` message so it's not obvious to me what this is doing by sending a blank response.  Is it just to avoid the error messages that happen when the `TALKREQ` handler in discv5 times out?

acolytec3 (approved):
This looks great overall.  Left one comment that I'd like clarity on but we can address that in days to come.  It looks great overall though.

PR #32, opened 2021-12-20
Small Changes to Browser Client Layout

ScottyPoi:
Local Node info turned into a Modal
Log moved to left column


acolytec3 (approved):
LGTM!  I fixed the random node lookup button by pointing it to the existing discv5 `findRandomNode` method.  Not sure why it wasn't working with `findNode` where you provided the input before but it works now.

Issue #36, opened 2021-12-21
Add JSON RPC interface to CLI client


acolytec3:
@emersonmacro Here's an issue to track your RPC work against

Below are two useful methods from the standard interface we would want to expose though feel free to add some other generic ones (maybe something like `addEnr`
`eth_getBlockByHash` - should take the standard parameters from the JSON RPC interface
`debug_getBlockRlp` - same
addEnr - just take something like a hex string as input

For now, you can just set up the rpc to start when the client starts and then just dump the parameters to the console.  We can implement the wiring later that actually ties it to the portal network machinery.

emersonmacro:
@acolytec3 Created a PR with some very basic scaffolding. It basically adds some CLI options, stands up an RPC server, and listens for one method (where all it does is dump params to console). I can build it out more from here over the next week or two.

emersonmacro:
@acolytec3 Thanks for merging 😄. So my plan is to basically mirror a lot of the patterns used in EthJS, like each namespace/prefix getting its own module and an RPCManager that glues everything together. Sounds reasonable? If so, I can work on building out the namespaces with a better selection of endpoints, and then maybe get into actually wiring up some of the basic calls with real information.

acolytec3:
I think we need to take stock of what's all really on the table in the near future for the Portal Network and what the Ultralight client really supports at present and I wouldn't want you to spend a lot of time speccing out potential RPC messages that will never end up getting used since so much is still in flux with regard to the exact data points the Portal Network will ultimately implement.  Take a look at [this PR](https://github.com/ethereum/portal-network-specs/pull/64/files) .  This is the current state of what should be supported from an RPC standpoint though we're still a long way from there even being specs to support all of what's outlined here.  My instincts say if we start by adding complete parsing of parameters for the generic `eth_getBlockByHash`  method as currently defined in the Ethereum JSON-RPC spec and are able to wire that up to the not yet fully implemented distribution of block data around the History Subnetwork in a subsequent PR, we'll be on the cutting edge of what's needed for now.

emersonmacro:
Sounds good - I'll focus on `eth_getBlockByHash`

acolytec3:
Just merged your latest PR.  Thanks!  

[...]

PR #41, opened 2021-12-27
Basic rpc support

emersonmacro:
First steps toward #36


acolytec3 (approved):
Looks great!  I moved the console logs for the RPC to a dedicated `RPC` channel in the debug logger utility and renamed the method to `discv5_nodeInfo` since that's one we'll definitely want to expose but not sure about all the others.  Great first step!

PR #54, opened 2022-01-04
Replace BN.js. with bigints

acolytec3:
Replace bn.js with bigints everywhere except `discv5` (to maintain minimal changes from upstream dependency)

### DIFF
--- a/packages/portalnetwork/src/stateSubnetwork/routingTable.ts
+++ b/packages/portalnetwork/src/stateSubnetwork/routingTable.ts
@@ -1,5 +1,4 @@
-import { BN } from 'bn.js';
-import { ENR, KademliaRoutingTable, NodeId } from '@chainsafe/discv5'
+import { ENR, NodeId } from '@chainsafe/discv5'
 import { distance } from './util'
 import { PortalNetworkRoutingTable } from '../client';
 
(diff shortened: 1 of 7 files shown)


packages/portalnetwork/src/stateSubnetwork/routingTable.ts
```diff
@@ -17,9 +16,9 @@ export class StateNetworkRoutingTable extends PortalNetworkRoutingTable {
             results.push(...bucket.values());
         });
         results.sort((a, b) => {
-            const diff = distance(new BN(id, 16), new BN(a.nodeId, 16)).sub(distance(new BN(id, 16), new BN(b.nodeId, 16)));
-            if (diff.isNeg()) return -1;
-            if (diff.isZero()) return 0;
+            const diff = distance(BigInt(id), BigInt(a.nodeId)) - distance(BigInt(id), BigInt(b.nodeId));
+            if (diff < 0) return -1;
+            if (diff === 0n) return 0;
```
holgerd77, line 21: I guess Richard @ricmoo once mentioned that one should aways use the BigInt constructor and not the n appendix since `BigInt` can easier be polyfilled (by simply importing a library with the same name and behavior) while `n` is a syntactical change which can't be read at all on older system and would need some complicated parsing.

Not sure, do we generally want to follow this convention when introducing BigInt somewhere?

//cc @ryanio @jochem-brouwer

holgerd77: Again, not sure though, `ethereum-cryptography` is using the `n` syntax as well I just saw: https://github.com/ethereum/js-ethereum-cryptography/pull/18/files#diff-f5834a87a1206796fe7f1c16283aa9a35061845df4394ae7e4a41daa12c3a9cdR169

//cc @alcuadrado

alcuadrado: bigints are not pollyfillable, as you can't override operators in javascript. 

I guess what Richard was referring to is that if you are willing to make an extra effort you can fall back to an alternative logic that doesn't use bigints as long as you don't use the new syntax (i.e. avoiding a syntax error).

I don't think that approach is worth it for a new project, especially a somewhat experimental one like this one.

holgerd77: > bigints are not pollyfillable, as you can't override operators in javascript.

Ah, that makes sense, thanks! 🙂

PR #66, opened 2022-01-09
Rpc manager

emersonmacro:
Moves the RPC methods into a simple class that also takes a `portalnetwork` client instance. The RPC methods have not actually been connected to the client yet.


acolytec3:
Looks good so far!  I fixed an export issue with portalnetwork and added logic to look in the local db for the requested block and return the rlp encoded blockheader if it exists.  This is just a first step but we'll need to add the appropriate logic with `sendFindContent` to retrieve the header from the network and then return it in the correct format.  I can't remember off hand if an RPC response is an RLP encoded block or the decoded block instead.

packages/cli/src/rpc.ts
```diff
@@ -1,17 +1,38 @@
...
+      const [blockHash, includeTransactions] = params
+      log(
+        `eth_getBlockByHash request received. blockHash: ${blockHash} includeTransactions: ${includeTransactions}`
+      )
+      const lookupKey = getContentId({ chainId: 1, blockHash: fromHex(blockHash.slice(2)) }, 0)
+      const value = await this._client.db.get(lookupKey)
+      if (value) {
+        return value
+      }
+      // TODO: Add sendFindContent to this to retrieve from network if not found in local DB
```
acolytec3, line 24: Lets maybe finish this `sendFindContent` piece for now where we retrieve the block header from the network and then return it in the correct format (either RLP or JSON blob, whichever is according to the JSON-RPC spec).  We can add the block bodies part later.

acolytec3 (changes requested):

packages/cli/src/rpc.ts
emersonmacro: Sounds good, will keep working on it. Just checked the specs and the content response should be RLP - https://github.com/ethereum/portal-network-specs/blob/master/history-network.md?plain=1#L61

acolytec3: For sure in the sendFindContent response.  This is actually what I was talking about - https://eth.wiki/json-rpc/API#eth_getBlockByHash  We need to mirror the standard Ethereum JSON RPC responses for methods with the same names.  Think of this RPC as a drop-in replacement for a full node in the set of method we are aiming to support.

acolytec3 (approved):
I'm going to go ahead and approve/merge this as I need the RPC manager method you added here for some other work I'm doing with the RPC.

PR #67, opened 2022-01-10
Bulk content update

acolytec3:
- Adds a bulk content loader to the browser client that loads nearest 100 blocks from a JSON blob of the format:
```json
{
  "0x0c1cf9b3d4aa3e20e12b355416a4e3202da53f54eaaafc882a7644e3e68127ec": {
    "rlp": "0xf9028ef90217a08faf8b77f...",
    "number": 200002
  },
...
}
```


acolytec3:
@emersonmacro @ScottyPoi I'm not 100% sure I'm satisfied with this approach to content loading.  It "works" from a browser client perspective but would be curious if you have thoughts on a better way to do it.  Basically, you click a button and asks you for a file like the ones in this dropbox - https://www.dropbox.com/s/y5n36ztppltgs7x/mainnetMM.zip?dl=0 which it then parses, sorts by distance from the ndoe, and loads the first 100.  Does this feel like the right approach?

packages/portalnetwork/src/historySubnetwork/types.ts
```diff
@@ -1,5 +1,10 @@
 import { ByteVectorType, ContainerType, NumberUintType, Union, UnionType } from '@chainsafe/ssz'
 
+/**
+ * @property chainId - decimal number representing the chain ID (e.g. Ethereum Mainnet is 1)
```
emersonmacro, line 4: minor feedback - `integer` or `integer number` instead of `decimal number`?

emersonmacro (approved):
I think this approach works. I could see it being extended with an input to select the number of items to import (100, 1000, all, etc). Probably makes sense to have a CLI-native way to do bulk content imports as well, maybe via a local script.

PR #77, opened 2022-01-13
Proxy updates

acolytec3:
## Proxy updates
- Adds new options for proxy to allow websocket server to run on localhost or LAN IP 
- Add option to specify ENR with public IP (taken from STUN server)

## Browser Client updates
- Adds new options to "Local Node Management" to set proxy address (defaults to `localhost`)
- Adds new start/stop node functionality

## Client updates
- Updates args parsing to correctly start proxy with localhost setting

### DIFF
--- a/packages/proxy/src/index.ts
+++ b/packages/proxy/src/index.ts
@@ -1,18 +1,27 @@
 import WS from 'ws'
 import * as dgram from 'dgram'
 import ipCodec from '@leichtgewicht/ip-codec'
+import yargs from 'yargs'
+import { hideBin } from 'yargs/helpers'
 const stun = require('stun')
 
 // TODO - replace console logs with debug logger
 
 const MAX_PACKET_SIZE = 1280
 
-const ws = new WS.Server({ host: '127.0.0.1', port: 5050, clientTracking: true })
+const servers: WS.Server[] = []
 
-const main = async () => {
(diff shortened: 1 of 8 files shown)


acolytec3:
@ScottyPoi @emersonmacro  would appreciate a review of this before I merge to make sure it's still usable.  These updates will make interop testing with local trin nodes much easier

packages/proxy/src/index.ts
```diff
@@ -71,12 +86,35 @@ const main = async () => {
...
+      break
+    case 'lan':
+      {
+        if (!args.ip) {
+          console.error('Must provide IP address for LAN option')
+          console.log('Exiting...')
+          process.exit(1)
+        }
+        const ws = new WS.Server({ host: args.ip, port: 5050, clientTracking: true })
+        startServer(ws, args.nat.includes('extip'))
```
emersonmacro, line 116: missing `servers.push(ws)`?

emersonmacro (approved):
left one question in `packages/proxy/src/index` but otherwise changes look good to me

ScottyPoi (approved):

packages/proxy/src/index.ts
acolytec3: Good catch!!!

PR #88, opened 2022-01-21
Display Block Data Component

ScottyPoi:
Basic component to display block data when it is added to db


acolytec3 (approved):

packages/browser-client/src/Components/ShowBlockContent.tsx
```diff
@@ -0,0 +1,154 @@
...
+      difficulty: Number(this.difficulty),
+      number: Number(this.number),
+      gasLimit: Number(this.gasLimit),
+      transactions: this.transactions,
+      timeStamp: this.timeStamp,
+      extraData: this.extraData,
+      mixHash: this.mixHash,
+      nonce: Number(this.nonce),
+    }
+  }
```
holgerd77, line 73: That would actually be a great use case for our @ethereumjs/block library. 🙂 (and I guess it would also make sense, since it would make things a lot more resilient in times of value verification and different types (pre-1559, post-1559 e.g.) blocks.

No sure if you want to introduce this dependency here though.

acolytec3: Good point.  We've already got the `block` dependency in the monorepo so maybe we can switch out `ethers` for `block` and make this one last dependency.

ScottyPoi: okay thanks :)

packages/browser-client/src/Components/ShowBlockContent.tsx
```diff
@@ -0,0 +1,154 @@
...
+} from '@chakra-ui/react'
+import { ethers } from 'ethers'
+import { PortalNetwork } from 'portalnetwork'
+import { useEffect, useState } from 'react'
+
+interface ShowBlockContentProps {
+  portal: PortalNetwork
+}
+
+class Block {
```
acolytec3, line 23: Lets switch this whole class + `ethers` out for the `@ethereumjs/block` dependency.  We can use the [`BlockHeader.fromRlpSerializedHeader`](https://github.com/ethereumjs/ethereumjs-monorepo/blob/23a7f35c471d7bc3797b3452d9be0947000b6102/packages/block/src/header.ts#L128) static constructor to get all of this data and don't have to duplicate it here.

acolytec3 (changes requested):
Per @holgerd77 's comment, lets move from Ethers to @ethereumjs/block

PR #89, opened 2022-01-26
Scotty poi/issue75

ScottyPoi:
Writing socket should now keep track of the seqNr's sent and AckNr's received, and compare the lists.


acolytec3 (approved):
LGTM!

PR #90, opened 2022-01-26
ci: add concurrency

ryanio:
I just added this to the monorepo (https://github.com/ethereumjs/ethereumjs-monorepo/pull/1667) and thought it would be nice to add to the workflow file here.

Not so important today since you only have one pretty speedy workflow file, but when additional ones are added if these lines are included you'll get the benefits of cancelling old runs on new commits to PRs.


acolytec3 (approved):
Awesome, thanks for dropping this in!

PR #92, opened 2022-01-27
Add new integration test CI

acolytec3:
- Builds new integration test that starts two nodes, plays PING/PONG, and then does OFFER/ACCEPT and verifies data transferred over uTP
- Simplifies data compilation when uTP stream completes


holgerd77:
Pretty cool! 👍 🎉

acolytec3:
> Pretty cool! 👍 🎉

Agree.  If I can get it to work across all our major functionality, it'll be a big win because I seem to keep breaking things when I add new features/rework stuff.

ScottyPoi (approved):

packages/portalnetwork/package.json
```diff
@@ -9,6 +9,7 @@
     "docs": "typedoc src/index.ts",
     "tape": "tape -r  ts-node/register",
     "test": "tape -r  ts-node/register 'test/!(integration)/**/*.spec.ts'",
+    "test:imtegration": "tape -r ts-node/register 'test/integration/**/*.spec.ts'",
```
ryanio, line 12: noticed this typo `test:imtegration` 😄 funny enough the workflow file doesn't use it `- run: npm run tape -- 'test/integration/wire.spec.ts'` so could be updated to `- run: npm run test:integration`

ryanio: also you don't really need a whole new workflow file, an additional job named `test-portalnetwork-integration` in the existing workflow file would have been fine (you could even rename the existing job to `test-portalnetwork-unit), and it would group the checks together as well, sorry just thought of this!

acolytec3: Good point. I hadn't thought about that but will definitely simplify the current test setup.  I'll fix my spelling errors too. Planning a cleanup PR for today to take care of several things.

PR #94, opened 2022-01-29
Block explorer:  Display Block data from findContent

ScottyPoi:
when portal client receives a new piece of block content it will display the block as a tabbed table.  

the parent hash hex is a link that will trigger find content for that block


packages/browser-client/src/Components/DisplayBlock.tsx
```diff
@@ -0,0 +1,107 @@
...
+  Tabs,
+  Table,
+  Thead,
+  Tr,
+  Th,
+  Td,
+  Text,
+  Link,
+} from '@chakra-ui/react'
+// eslint-disable-next-line implicit-dependencies/no-implicit
```
ScottyPoi, line 17: I don't know why I'm getting this error --
It compiles fine if I disable it like this

acolytec3 (approved):
LGTM!  Thanks for changing to the ethjs libraries/

packages/browser-client/src/Components/DisplayBlock.tsx
```diff
@@ -0,0 +1,104 @@
...
+    mixHash: header.mixHash!,
+    nonce: header.nonce!,
+    number: Number(header.number!),
+    parentHash: header.parentHash!,
+    receiptTrie: header.receiptTrie!,
+    stateRoot: header.stateRoot!,
+    timestamp: new Date(parseInt(header.timestamp!)).toString(),
+    transactionsTrie: header.transactionsTrie!,
+    uncleHash: header.uncleHash!,
+  }
```
holgerd77, line 53: You should really also remove this `headerObj` structure, even if you have somewhat of an implementation reason for it (the non-null assertion operator?).

Beside the fact that it looks heavily redundant, this won't work for old-style blocks pre-London not including the 1559 fee mechanism yet and therefore not having a `baseFee` and `baseFeePerGas` field. So this won't work for the chain history network to full extend yet and is therefore unnecessarily limiting the scope here.

holgerd77: (otherwise: nice to see the EthereumJS libs integrated here 🙂)

ryanio: agree with holger that it should be more dynamic to be more resilient for changes from hardforks

for the returned type if you needed the non-null assertions you can try casting as [`Required<Type>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype) so `Required<JsonHeader>`

Issue #52, opened 2022-01-03
Abstract Content Lookup into its own Class


acolytec3:
The `FINDCONTENT` handler flow currently just sends a FINDCONTENT request to a given node and then recursively sends the same requests to any ENRs it receives until it either finds the sought content or it runs out of nodes to query.  When content is successfully retrieved, the history network spec calls for OFFERing that content to the nearest node that is interested in the content (i.e. it's within its radius) that didn't respond with the actual content.  This will probably require us to abstract out the FINDCONTENT process into it's own class or somehow track open content lookups that are in process so we can identify the nearest node that didn't have the content that should.

acolytec3:
#116 and #117 allow us to spin up ephemeral devnets of any arbitrary size.  I've been able to see nodes find content that is two hops away so that's promising though we need to look more at the algorithm and see where it varies from the Kademlia lookup algorithm since the portal network specs say we should be following kademlia here and I just implemented a naive starting point to give us something to work with.

acolytec3:
Going to close this one in favor of opening a new issue for improving the lookup mechanisms.

PR #102, opened 2022-02-08
Portal network changes for content lookup

acolytec3:
Adds a new `lookup` class for content lookups in the `portalnetwork` module.


ScottyPoi:
is this instead of #52 or part of it?

acolytec3:
I'll need to solve for #52 in this PR.  Thanks for the reminder.

Issue #105, opened 2022-02-09
Browser Client clean-up


acolytec3:
- [ ] Now that the `Find Content` button hooks into `contentLookup` instead of `sendFindContent`, we should move it out of the `AddressBookManager` into its own component
- [ ] The `DisplayBlock` breaks when you load a large amount of content with the `Load Content` button and only displays one block. The dropdown should be fixed to show all the blocks currently available in the DB.

acolytec3:
Fixed by #111

PR #111, opened 2022-02-15
FindConent Component

ScottyPoi:
#105 

* Separated Address Book Manager and Content Manager into separate components
* Altered node communication controls to utilize a select menu of nodes in address book
    * One set of buttons instead of new set for each node 
* Combined Content Manager with Select Block menu for block
* Updated App layout to accommodate changes


acolytec3 (approved):
LGTM

PR #119, opened 2022-02-18
Add basic prometheus metrics reporting

acolytec3:
Adds the most basic of prometheus metrics reporting from a running client (just number of peers connected for both discv5 and history network routing table)


holgerd77:
Oh, this Prometheus thing looks really interesting, never stumbled upon that.

(for reference: https://prometheus.io)

PR #125, opened 2022-02-22
Insert 5 digits of NodeId as prefix for debug logs

ScottyPoi:
For the purpose of distinguishing the different logs of many nodes running in one console.  Each unique log name will be colored uniquely, and the nodeId prefix will allow for easier reading of the debug printout.


acolytec3 (approved):
LGTM!

PR #138, opened 2022-02-25
General routing table management

### LINKED ISSUE #129: Generalize all subnetwork handling
Currently, the entire `portalnetwork` module assumes that we are operating on the History Network only.  This was done for easy of implementation in the early days and has never been resolved.  This is woven throughout the stack and really needs to be generalized so that all messages/table lookups/etc take a SubnetworkId as an input and query the correct routing table or send a message with the correct Subnetwork Id that is applicable.  One good example is [here](https://github.com/ethereumjs/ultralight/blob/662fda93b4797d028564554bad07473321c80728/packages/portalnetwork/src/client/client.ts#L833) where every single Portal Network message that is sent is checked against the History Network Routing Table regardless of the intended subnetwork and the clunky implementation of [`updateRoutingTable`]([private](https://github.com/ethereumjs/ultralight/blob/662fda93b4797d028564554bad07473321c80728/packages/portalnetwork/src/client/client.ts#L767)).  A good solution will not just implement a switch statement in every method that checks the correct routing table based on the `SubnetworkID` but gives us a coherent way of referencing the correct routing table through a simple lookup, something like `this.routingtables[subnetworkId]` so could be as simple as implementing a new array of routing tables at the top level `portalnetwork` class though this could be tricky since subnetwork routing table implementations are not identical (notably the State Network).

acolytec3:
Mostly fixes #129 

The outstanding items lie in how to handle utp socket setup when we don't have a reference point for the subnetwork the content is being streamed for when setting up the uTP socket.

### DIFF
--- a/packages/portalnetwork/test/integration/wire.spec.ts
+++ b/packages/portalnetwork/test/integration/wire.spec.ts
@@ -67,10 +67,9 @@ tape('Portal Wire Spec Testing', async (t) => {
         portal1.client.once('multiaddrUpdated', () => portal2.start())
 
         portal2.client.once('multiaddrUpdated', async () => {
-          portal2.historyNetworkRoutingTable.insertOrUpdate(
-            portal1.client.enr,
-            EntryStatus.Connected
-          )
+          portal2.routingTables
+            .get(SubNetworkIds.HistoryNetwork)!
+            .insertOrUpdate(portal1.client.enr, EntryStatus.Connected)
           const res = await portal2.sendPing(
             portal1.client.enr.nodeId,
             SubNetworkIds.HistoryNetwork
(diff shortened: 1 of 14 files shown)


packages/portalnetwork/test/integration/wire.spec.ts
```diff
@@ -90,6 +88,7 @@ tape('Portal Wire Spec Testing', async (t) => {
     const child = spawn(process.execPath, [file])
     let portal1: PortalNetwork
     let portal2: PortalNetwork
+    tape.onFailure(() => end(child, [portal1, portal2], st))
```
acolytec3, line 91: @ryanio I discovered this `tape.onFailure()` method while triaging a hanging integration test for this PR.  When you combine it with a `timeout` in an async test, it makes for a nice way to clean-up things that cause tests to hang so they still terminate if there's a failure.  Could be a helpful additional tool in cleaning up tests in the `client ` if we identify new race conditions/tests that occasionally fail due to open handles.

ScottyPoi (approved):

packages/portalnetwork/test/integration/wire.spec.ts
ryanio: cool, thanks for the tip!

PR #146, opened 2022-03-02
Utp socket content type

ScottyPoi:
* Adds Content type to utp_socket metadata.
* updates utp protocol to create typed sockets
* updates utp method names to mirror portal network message names
* updates utp reader to emit stream event with content type
* updates client to track content types
* updates client to use new utp method names.
* updates streamed content event to contain content type
----
Also adds tests to seeder script to aid in refactoring uTP

### DIFF
--- a/packages/portalnetwork/src/wire/utp/Protocol/utp_protocol.ts
+++ b/packages/portalnetwork/src/wire/utp/Protocol/utp_protocol.ts
@@ -1,10 +1,16 @@
 import { _UTPSocket } from '../Socket/_UTPSocket'
-import { Packet, PacketType } from '..'
+import { ConnectionState, Packet, PacketType } from '..'
 import { debug, Debugger } from 'debug'
 import { PortalNetwork, SubNetworkIds } from '../../..'
 import { Discv5 } from '@chainsafe/discv5'
 import { fromHexString } from '@chainsafe/ssz'
 import { serializedContentKeyToContentId } from '../../../util'
+import { HistoryNetworkContentTypes } from '../../../historySubnetwork/types'
+import { check } from 'prettier'
+
+const testArray = fromHexString(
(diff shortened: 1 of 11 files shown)


packages/portalnetwork/src/wire/utp/Protocol/utp_protocol.ts
```diff
@@ -52,68 +58,122 @@ export class UtpProtocol {
...
+    contentTypes: number[],
     networkId: SubNetworkIds
   ) {
-    const socket = new _UTPSocket(this, remoteAddr, 'writing', networkId)
-    const value = await this.portal.db.get(serializedContentKeyToContentId(contentKeys[0]))
-    socket.content = fromHexString(value)
-    this.sockets[remoteAddr + connectionId] = socket
-    await this.sockets[remoteAddr + connectionId].sendSynPacket(connectionId)
+    const socketKey = this.getSocketKey(remoteAddr, connectionId)
+    contentKeys.forEach(async (content, idx) => {
```
acolytec3, line 86: Does this `forEach` imply we're creating multiple sockets (one per `contentType`) but that share the same `socketKey`?  It's not 100% clear to me why we're passing the `desiredContentTypes` as an array here and if this is attempting to create multiple sockets or not (since it doesn't look like it will since).  If we're using only a single connectionId, our current use case for uTP in the `OFFER`/`ACCEPT` flow implies we'll only send a single piece of content over the uTP socket (corresponding to a single `contentKey`).  And we can derive the `contentType` for the socket from the `contentKey`, right?

ScottyPoi: Yes to the 'single piece of content' per socket thing.  For each piece of accepted content it will make a new socket.  And yes, sharing the socketKey is an unsolved problem that arises from that.  I see what you mean about the socket deriving the contentType for itself.  Yes, that would be simpler than saving the array of types.

packages/portalnetwork/src/wire/utp/Socket/setup.md
```diff
@@ -0,0 +1,60 @@
...
+    - B sends ACK packet with seqNr: RANDB+1, ackNr: 2
+    - A sends DATA packet with seqNr: 3, ackNr: RANDB+1
+    - B sends ACK packet with seqNr: RANDB+2, ackNr: 3
+    - A sends FIN packet with seqNr: 4, ackNr: RANDB+2
+    - B checks to see if it has packets #1, #2, #139
+      - sends ACK packet with seqNr: RANDB+3, ackNr: 4
+      - Reader recompiles, sends Stream event to client
+      - B closes socket and opens new Reading socket
+    - A receives STATE with ackNr: FIN (4), closes socket
+    - Repeat
```
holgerd77, line 60: That's a great (or maybe: the best) way of exposing the work you guys are doing, just writing good documentation! 👍 😄 🎉

PR #147, opened 2022-03-02
Add-some-tests

ScottyPoi:
update both seeder script and tape tests to run utp tests


ScottyPoi:
Not specifically related to utp_socket content type issue.  seeder script will run portal content tests and expose uTP bugs

PR #153, opened 2022-03-09
Refactored uTP

ScottyPoi:
WIP

Still requires testing for OFFER/ACCEPT.

Refactored uTP works by creating "request" class.

This request class separates actual portal network info from uTP.
Also allows for content to be resent if socket fails.
Keeps the requested content connected to the contentKey for the db, the type of content, subnetwork id, and peerId
Request and sockets can be destroyed without uTP connection being destroyed.

- PortalNetwork client constructor now creates a "PortalNetworkUtp" class stored as this.utp and passes itself to the PortalNetworkUtp constructor.
- On the client side, when a need for utp arises from FINDCONTENT/FOUNDCONTENT or OFFER/ACCEPT, a HistoryNetworkContentRequest is created with this.utp.handleNewHistoryNetworkRequest
- A new request is created with an array of contentKeys, a peerId, uTP id, requestCode, and an array of contents (if sending).
- Request Code refers to one of four distinct PortalNetworkUtp signal flows.
0:   FINDCONTENT_WRITE
1:   FINDCONTENT_READ
 2:  OFFERACCEPT_WRITE
 3:  OFFERACCEPT_READ
- The PortalNetworkUtp keeps a record of open requests identified by a key derived from the utp connection ids for that transaction
- The UtpSocket for the transaction will be created with the Request.  Sockets can be opened and closed by PortalNetworkUtp while the request persists.  This helps in managing failures as well as multiple pieces of content sent through offer/accept.
- PortalNetworkUtp handles the serialized packets that arrive via TALKREQ messages.  Depending on the Request-Code associated with that connection, the deserialized packet will be handled in 4 different ways.
- BasicUtp and PacketSenders are agnostic layers of non-PortalNetwork uTP logic that live between PortalNetworkUtp and the UtpSocket.  This is where packets are read, written, created, or sent.
- A ContentReader or ContentWriter is created when the sequence numbers have been established.  (either the 2nd or 3rd packet).
- ContentWriter chunks the content into a predetermined size, and stores them with the calculated seqNr's as keys.  
- When the DATA phase of uTP is triggered, the ContentWriter will manage the sending of DATA packets as well as the tracking of ACK packets.
- ContentReader keeps a record of incoming DATA packet content, with the packet's seqNr as a key.  
- When all DATA is sent, a FIN packet is sent, signaling to the reading socket what the final seqNr will be.  When that packet has been Acked, the socket will close.  
- When ContentReader sees a FIN packet, it verifies that no packets are missing, and then recompiles the data back together.  It will pass the compiled data back through the Request to PortalNetworkUtp, which will add it to the Database using the contentKey data also stored in the request.
- Reading socket will ACK the FIN packet, and close.
- If there is more content to be sent in that Request, new sockets will be opened for the same request, and then rinse and repeat.  If there is no more content to be sent, the whole request can be deleted.


packages/portalnetwork/src/wire/utp/PortalNetworkUtp/PortalNetworkUTP.ts
```diff
@@ -0,0 +1,495 @@
...
+          undefined,
+          'read',
+          this.logger
+        )
+        return socket
+    }
+  }
+
+  async handleUtpPacket(packetBuffer: Buffer, srcId: string): Promise<void> {
+    const requestKey = this.getRequestKeyFromPortalMessage(packetBuffer, srcId)
```
acolytec3, line 241: What happens if we don't have a request key?  I know the `getRequestKeyFromPortalMessage` logs an error but should we do anything else (e.g. send an empty talkresp - or is that already happening)?

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/PortalNetworkUTP.ts
```diff
@@ -0,0 +1,495 @@
...
+    if (this.openHistoryNetworkRequests[keyA] !== undefined) {
+      return keyA
+    } else if (this.openHistoryNetworkRequests[keyB] !== undefined) {
+      return keyB
+    } else if (this.openHistoryNetworkRequests[keyC] !== undefined) {
+      return keyC
+    } else if (this.openHistoryNetworkRequests[keyD] !== undefined) {
+      return keyD
+    } else {
+      this.logger(`Cannot Find Open Request for socketKey ${keyA} or ${keyB} or ${keyC} of ${keyD}`)
```
acolytec3, line 298:
```suggestion
      this.logger(`Cannot Find Open Request for socketKey ${keyA} or ${keyB} or ${keyC} or ${keyD}`)
```

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/PortalNetworkUTP.ts
```diff
@@ -0,0 +1,495 @@
...
+import { BasicUtp } from '../Protocol/BasicUtp'
+import { HistoryNetworkContentRequest } from './HistoryNetworkContentRequest'
+
+type UtpSocketKey = string
+
+export enum RequestCode {
+  FOUNDNDCONTENT_WRITE = 0,
+  FINDCONTENT_READ = 1,
+  OFFER_WRITE = 2,
+  ACCECPT_READ = 3,
```
acolytec3, line 17:
```suggestion
  ACCEPT_READ = 3,
```
Also fix the places this enum is used.

acolytec3 (changes requested):
Mostly niggly feedback.  I need to do some more review but this is a start.

packages/portalnetwork/src/client/client.ts
```diff
@@ -577,18 +612,31 @@ export class PortalNetwork extends (EventEmitter as { new (): PortalNetworkEvent
...
-    const header = BlockHeader.fromRLPSerializedHeader(Buffer.from(content))
-    this.addContentToHistory(
-      1,
-      HistoryNetworkContentTypes.BlockHeader,
-      toHexString(header.hash()),
-      content
-    )
+    if (contentType === HistoryNetworkContentTypes.BlockHeader) {
+      this.addContentToHistory(
+        1,
```
acolytec3, line 627:
```suggestion
        0,
```
Content type 1 is blockbody

packages/portalnetwork/test/wire/utp.spec.ts
```diff
@@ -140,3 +197,40 @@ tape('uTP encoding tests', (t) => {
...
+  )
+  await portal.start()
+  portal.client.addEnr(cli_enr)
+  // const syn = await portal.uTP.initiateConnectionRequest(cli_nodeId, 5555)
+  t.test('Portal Client Test', async (st) => {
+    st.ok(portal.client.isStarted(), 'Portal Client Started')
+    const pong = await portal.sendPing(cli_nodeId, SubNetworkIds.HistoryNetwork)
+    st.ok(pong, 'Ping/Pong 1 successful')
+    const res = await portal.historyNetworkContentLookup(0, contentKey)
+    st.ok(
```
acolytec3, line 230: Lets change this test to use `st.equal...`

packages/portalnetwork/test/wire/utp.spec.ts
```diff
@@ -140,3 +197,40 @@ tape('uTP encoding tests', (t) => {
     st.end()
   })
 })
+
+tape('uTP packet handling', async (t) => {
```
acolytec3, line 201: I think this whole test should already be covered by our integration test that is intended test uTP.  I ran that test locally and it looks like it's broken right now will need to look at that.

acolytec3 (approved):
LGTM (once we fix the tests!)

PR #159, opened 2022-03-15
Selective ack handling and testing

ScottyPoi:
1.
Handling of out of order data packets with selective ack implemented.

2. Handling of selective ack implemented as far as ack tracking.  No fallback yet for packets that never arrive.  

3. Disabled by default -- but tested successfully.
There is a version of uTP writer called BadWriter.  This will send all packets, but out of order.  uTP reader has proven to be robust against this type of failure.


acolytec3:
I will confess that I'm not tracking exactly how the badWriter gets called in your tests but everything looks good here and I've fixed several logic bugs in the table maintenance code that I've identified that should smooth out some of our disconnect issues so going to go ahead and merge.

acolytec3 (approved):

PR #164, opened 2022-03-18
More utp tweaks

acolytec3:
Additional uTP tweaks to match Fluffy implementation for `sendFindContent`


acolytec3:
I'm not 100% sure these all make sense but they do allow our utp code too work with fluffy in the findContent scenario so now utp should work in all contexts for a single piece of content.

PR #166, opened 2022-03-23
update history network rpc methods

ScottyPoi:
portal_history_findContent
portal_history_offer

### DIFF
--- a/packages/cli/src/rpc.ts
+++ b/packages/cli/src/rpc.ts
@@ -97,22 +97,6 @@ export class RPCManager {
       this.log(`response received to findNodes ${res?.toString()}`)
       return `${res?.total ?? 0} nodes returned`
     },
-    portal_offer: async (params: [string, string, number]) => {
-      const [dstId, blockHash, contentType] = params
-      if (!isValidId(dstId) || contentType < 0 || contentType > 2) {
-        return 'invalid parameters'
-      }
-      const contentKey = HistoryNetworkContentKeyUnionType.serialize({
-        selector: contentType,
-        value: {
-          chainId: 1,
-          blockHash: fromHex(blockHash.slice(2)),
-        },
-      })
-      const res = await this._client.sendOffer(dstId, [contentKey], SubNetworkIds.HistoryNetwork)
-      this.log(`response received to offer ${res?.toString()}`)
-      return `${shortId(dstId)} ${res ? 'accepted' : 'rejected'} offer`
-    },
     portal_ping: async (params: [string]) => {
       const [enr] = params
       const encodedENR = ENR.decodeTxt(enr)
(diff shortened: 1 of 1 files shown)


packages/cli/src/rpc.ts
```diff
@@ -121,6 +105,32 @@ export class RPCManager {
       this.log(`TEST PONG received from ${encodedENR.nodeId}`)
       return `PING/PONG successful with ${encodedENR.nodeId}`
     },
+    portal_history_findContent: async (params: [string, Uint8Array]) => {
+      const [enr, contentKey] = params
+      await this._client.sendFindContent(enr, contentKey, SubNetworkIds.HistoryNetwork)
```
acolytec3, line 110:
```suggestion
     const res =  await this._client.sendFindContent(enr, contentKey, SubNetworkIds.HistoryNetwork)
     return res
```
Can we do something like this to return the response from this message?  I don't remember exactly what it looks like so we may need to decode it or something but we should always return some sort of response from RPC methods (even if it's only `request succeeded` or something like that)

packages/cli/src/rpc.ts
```diff
@@ -121,6 +105,32 @@ export class RPCManager {
...
+      const contentKeys = blockHashes.map((blockHash, idx) => {
+        return HistoryNetworkContentKeyUnionType.serialize({
+          selector: contentTypes[idx],
+          value: {
+            chainId: 1,
+            blockHash: fromHex(blockHash.slice(2)),
+          },
+        })
+      })
+      await this._client.sendOffer(dstId, contentKeys, SubNetworkIds.HistoryNetwork)
```
acolytec3, line 132: Here again, let's return something from this RPC call (presumably the contents of the ACCEPT response or else an empty array if we don't get anything)

acolytec3:
Thanks for starting on this!  I don't know if you have opinions about what we should return in these RPC calls.  The draft specs all just say to "return the message ID" but we don't expose the actual TalkReq message ID in the `portalnetwork` API so we'd have to rework our module to do so.  I guess a lot of it depends on how much work we're going to put into the RPC endpoints.  They certainly have their place but I'm not sure we need to recreate the entire RPC spec that's in draft form [here](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ogenev/portal-network-specs/jsonrpc-assembled-specs/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false) given that Ultralight isn't ultimately intended to be run on NodeJS (or at least our CLI client isn't intended to be a production app longer term).

PR #171, opened 2022-03-25
Browser redesign

ScottyPoi:
Migrating back over from the fork I made to work on this.
Redesign of Browser-Client into "Ultralight Portal Network Explorer"
Still capable of manual 'send ping', 'find nodes' and 'send offer'

### DIFF
--- a/packages/browser-client/README.md
+++ b/packages/browser-client/README.md
@@ -1,6 +1,6 @@
 # Ultralight Block Explorer
 
-This is a proof of concept that will start an Ultralight portal network client in the browser that leverages a UDP proxy service to connect to other nodes in the network,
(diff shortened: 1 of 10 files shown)


acolytec3:
This looks awesome in general.  I realized after I made the commits that I think I broke your pretty chart a bit.  I'm not 100% sure what was going on with your distance calculation but it was different than the actual `log2Distance` that Kademlia provides and it's easier in general to reason about the log2Distance since the nodes are already placed into those buckets in the routing table.  Is it difficult to update your table to just use the range of log2Distances of the `sortedDistList` nodes for the pink lines indicating the distances?  

There's also something goofy going on with the `parentSearch` thing where it doesn't retrieve the block the first time that I'll look into some more but awesome work overall!

PR #173, opened 2022-03-28
Tweaks to network visual

ScottyPoi:
Ongoing development of visualization components

### DIFF
--- a/packages/browser-client/src/App.tsx
+++ b/packages/browser-client/src/App.tsx
@@ -32,7 +32,7 @@ import {
   TableCaption,
   Center,
 } from '@chakra-ui/react'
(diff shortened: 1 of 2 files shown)


acolytec3:
Is this ready for review? From looking at the code, it appears the network diagram is now showing a distribution relative to the blockhash currently being searched for, right?

ScottyPoi:
just playing with ideas, trying to show where the content is coming from.  you see a black marker at the blockhash address and the "closest" node lights up yellow.  i know the whole eyeball thing is rather pointless, but it has been interesting to make

ScottyPoi:
maybe just leave here in a PR for now while I keep messing with it

PR #177, opened 2022-03-29
WIP - Rework of dev tools drawer

acolytec3:
- Reworks dev tools drawer using `VStack`
- Adds "Load Content from File" option back


ScottyPoi (approved):
Kept all your changes.
Added a scroll to the peers menu

Issue #170, opened 2022-03-23
Add liveness checks for routing tables


acolytec3:
With [#168](https://github.com/ethereumjs/ultralight/issues/168), we can no longer count on the discv5 session liveness checks to help keep our routing tables fresh.  As such, we need to build out some sort of liveness check that regularly pings all the nodes in our routing table to make sure they're still online.

PR #178, opened 2022-04-02
Liveness Check

ScottyPoi:
Re: #170

Added private method client.livenessCheck() that pings all nodes in routing table to test connection

Added liveness check to the 'bucket refresh' method that is called every 30 seconds.


acolytec3 (approved):
LGTM!

PR #184, opened 2022-04-07
Initial setup for mobile app

acolytec3:
This is a very simple first start to build a mobile app version of the browser client.  It is a clone of the `browser-client` package with the additional scaffolding to build on Android.  I also made one minor change to `App.tsx` to use the mobile transport layer instead of websockets.  It's not perfect yet and hot reloading isn't working for some reason but if you follow the very minimal steps in the readme, notionally, you should be able to build and run the app on Android.


acolytec3:
> Great start partying_face, I guess it would be good to get this as little redundant as possible from the start, otherwise things get likely messy relatively quickly.
> 
> Any way to directly draw in the `browser-client` files with some auto-post-draw-in modifications? thinking I have no idea, just thinking loud. Maybe @ryanio can give some assistence on this if this is feasible? But let me openly know if this doesn't make sense in this stage of the work.

We definitely can do it all in one package though I will have to do some research on the best way to detect the build environment (maybe just through setting an environmental variable) so we use the right transport layer.  Let me experiment a little bit and see.  Capacitor may have a way to do this already.

acolytec3:
Okay, redid this with everything baked into the `browser-client` package.  Would be great if some brave soul tries to follow the steps I outlined in the `mobile` readme and tries to build the app and see if works.

ryanio:
great work! i was thinking, perhaps you want to make a new package for the android code like `packages/android`. the idea being the `browser-client` package should just be consumed as a dependency to keep the separation clean (saw that you deduped it already, nice!)

acolytec3:
I'll look at it some more. I've never consumed a react app as a dependency so will have to see if it makes sense.

ryanio:
Sounds good, thanks for considering, yeah no rush for this PR you can do in the future.

ScottyPoi (approved):
Additional UI improvements and bug fixes to be handled in different PR's.  Here we have a working way to demonstrate both browser and mobile apps.  :+1:

PR #188, opened 2022-04-10
Default-connections

ScottyPoi:
Fluffy Buttons will connect to a public fluffy bootnode.

Default contentKey will return a block from Fluffy node.

Default Start Node input for browser is "cloud proxy" IP.

ENR Copy button moved to Dev Tools

Footer links are active


acolytec3 (approved):
LGTM

Issue #183, opened 2022-04-06
Landing Page


ScottyPoi:
Rather than just the 'Start Node' component, the opening screen when you start the app could have content re: #182 and an evil eye of doom

PR #194, opened 2022-04-12
Hamburger-menu

ScottyPoi:
re: #183 

Top Left Menu Icon now has content.

* Basic app manual
* What is Portal Network? q&a
* Links to implementations

### DIFF
--- a/packages/browser-client/capacitor.config.ts
+++ b/packages/browser-client/capacitor.config.ts
@@ -15,7 +15,7 @@ switch (process.env.CAP_ENV) {
       ...baseConfig,
       webDir: 'public',
       server: {
-        url: 'http://[your webpack dev server address here]',
+        url: 'http://192.168.0.194:8080',
         cleartext: true
       }
     };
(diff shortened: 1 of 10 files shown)


acolytec3:
I think the content is great in this but I don't love the presentation.  Not a UI expert but tapping the hamburger menu to get "More Info" feels counter-intuitive to me. The content feels really crowded in the drawer component so maybe put it in a modal that overlays the whole screen when you click on a "More Info" option.  I think my preference would be something like embed both the dev tools and the "More Info" buttons or links in a hamburger menu and then open a modal that fills the screen if/when you tap on it and then show this content.  I'll play around with it a little bit this morning and see what I come up with.

acolytec3:
Here's what it looks like on mobile with my reworking.  It's not perfect but this feels a little more natural to me.  Thoughts?

Screenshot_20220412-092020_Ultralight
Screenshot_20220412-092016_Ultralight
Screenshot_20220412-092026_Ultralight

acolytec3:
@ScottyPoi Take a peek at my changes and see what you think.  It probably bears further refinement but this feels maybe less busy on the main screen to me.  The menu is certainly busy but maybe good enough for now?  It also handles the starting the proxy on web (and only shows that startnode option if it detects we're on the web platform).

ScottyPoi:
The only part that bugs me is that now you have to scroll to see everything in the menu, but i can live with that

PR #196, opened 2022-04-12
Just start browser node with public proxy

ScottyPoi:
Browser-client will start automatically with public proxy.  Bypassing START NODE step


acolytec3 (commented):
I think for this we need to put the "set proxy address" component inside dev tools or we lose the ability to do local development/testing.  Will incorporate this in my experimentation on the hamburger menu.

acolytec3:
I've picked up this change in the additional commits I made on #194

PR #197, opened 2022-04-12
Bootnodes-bootup

ScottyPoi:
App will connect to a list of bootnodes on startup

Bootnode buttons replaced with progress bar showing connection attempts

"CONNECT TO NODE"  function moved to dev tools

Spinners and Skeletons added to block explorer during retrieval time


acolytec3 (approved):
I like progress bars!

PR #198, opened 2022-04-12
UTP-fixes

ScottyPoi:
Hacky fix for weird uTP loop.

Also fixes a browser error related to "unique keys"

### DIFF
--- a/packages/portalnetwork/src/wire/utp/PortalNetworkUtp/HistoryNetworkContentRequest.ts
+++ b/packages/portalnetwork/src/wire/utp/PortalNetworkUtp/HistoryNetworkContentRequest.ts
@@ -47,13 +47,17 @@ export class HistoryNetworkContentRequest {
         await sendSynPacket(this.socket)
         break
       case RequestCode.OFFER_WRITE:
-        this.socket = this.sockets.pop()!
-        this.contentKey = this.contentKeys.pop()!
-        writer = await this.socket!.utp.createNewWriter(this.socket, 2)
-        this.writer = writer
-        await sendSynPacket(this.socket)
-        this.socket.state = ConnectionState.SynSent
-        break
+        if (this.sockets.length > 0 && this.sockets.length > 0) {
+          this.socket = this.sockets.pop()!
+          this.contentKey = this.contentKeys.pop()!
+          writer = await this.socket!.utp.createNewWriter(this.socket, 2)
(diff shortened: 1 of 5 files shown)


packages/portalnetwork/src/wire/utp/PortalNetworkUtp/HistoryNetworkContentRequest.ts
```diff
@@ -47,13 +47,17 @@ export class HistoryNetworkContentRequest {
...
         break
       case RequestCode.OFFER_WRITE:
-        this.socket = this.sockets.pop()!
-        this.contentKey = this.contentKeys.pop()!
-        writer = await this.socket!.utp.createNewWriter(this.socket, 2)
-        this.writer = writer
-        await sendSynPacket(this.socket)
-        this.socket.state = ConnectionState.SynSent
-        break
+        if (this.sockets.length > 0 && this.sockets.length > 0) {
```
acolytec3, line 50: Should this be?
```suggestion
        if (this.sockets.length > 0 && this.contentKeys.length > 0) {
```

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
```diff
@@ -198,6 +198,15 @@ export class UtpSocket extends EventEmitter {
     return this.readerContent
   }
 
+  async handleResetPacket() {
```
acolytec3, line 201: Will this socket get garbage collected at this point by the `uTpProtocol` class?

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/PortalNetworkUTP.ts
```diff
@@ -447,22 +447,8 @@ export class PortalNetworkUTP {
   }
   async handleResetPacket(request: HistoryNetworkContentRequest) {
     const requestCode = request.requestCode
+    await this.protocol.handleResetPacket(request.socket)
```
acolytec3, line 450: Do we need to do this step at all since the next line deletes the socket from the table of open sockets?

ScottyPoi:
I don't think any of this is necessary anymore with the gossip fixes.
Guarding against the infinite loop thing is a good idea, but i might look for a more elegant way to go about it.
Can table this for now i think

PR #206, opened 2022-04-15
Abstract discv5 transport layer

acolytec3:
- Rework transport layers as an interface object in discv5 transport layer
- Add necessary types from `discv5` in `portalnetwork`


ScottyPoi:
Nothing seems broken on the browser-client side.
CLI packages is failing to build though.

```
error TS2688: Cannot find type definition file for 'bcrypto'.
  The file is in the program because:
    Entry point of type library 'bcrypto' specified in compilerOptions
```

acolytec3:
> Nothing seems broken on the browser-client side. CLI packages is failing to build though.
> 
> ```
> error TS2688: Cannot find type definition file for 'bcrypto'.
>   The file is in the program because:
>     Entry point of type library 'bcrypto' specified in compilerOptions
> ```

Thanks!  This is resolved (at least for now though we may have to address again if/when we finally boot our discv5 fork out of the monorepo).  Going to go ahead and merge at this point since it looks like everything is working correctly.

Issue #205, opened 2022-04-14
Remove `discv5` fork from monorepo


acolytec3:
I discussed this with the maintainers of `@chainsafe/discv5` and we've agreed in principle to move our discv5 changes back upstream to their package so we can remove this package from our monorepo and have `portalnetwork` just depend on their upstream version.  We need to a line by line diff analysis between our current fork and the upstream master to identify any other changes but below are the current know deviations we need to pass back upstream:
- [x] Add a transport layer interface object to the `discv5` constructor to allow any transport layer that implements the `transportlayer` interface to be passed in - fixed by #206 
- [x] Update the TalkRequest and TalkResponse handlers to accept an ENR as an alternative to a `nodeId` - this is technically already done in our fork. Handled in the upstream PR
- [x] Determine if we really need that new session event I created `multiaddrUpdated` since we should theoretically be able to use just use the `addrVoting` process in `discv5` to identify our public IP and update accordingly.  This would simplify both the websocket transport implementation and the proxy if this works. - addressed in #207 
- [x] Remove customized debugging from our discv5 fork - this done in the PR submitted to the upstream `discv5` repo [here](https://github.com/ChainSafe/discv5/pull/169)
- [x] Merge upstream changes in `@chainsafe/discv5` - done [169](https://github.com/ChainSafe/discv5/pull/169)
- [x] Remove `discv5` package and replace dependencies in all other packages.  Finished in #229
cc: @dapplion @wemeetagain @ScottyPoi

acolytec3:
I've submitted an upstream PR the `@chainsafe/discv5` with our needed changes and have a branch working off a separated out `discv5` repo.  Once the upstream PR gets merged, will move this forward again.

acolytec3:
My upstream PR has been merged into `discv5`.  Once [this release](https://github.com/ChainSafe/discv5/pull/171) is published, we can finish this out!

acolytec3:
New discv5 release!  https://github.com/ChainSafe/discv5/releases/tag/v0.7.0  One more big push and we're done (hopefully!)

acolytec3:
Done and done.

dapplion:
:tada:

PR #207, opened 2022-04-15
Use addrVotes instead of proxy reflection to update ENR

acolytec3:
Removes public IP reflection from proxy and has `discv5` client use builtin `addrVotes` to update public IP instead
Also addresses point 3 of #205 by removing now unnecessary `ITransportEvents` in our `discv5` fork to make upstream merging easier.


acolytec3:
I've fixed the integration test on `portalnetwork`.  I think the last thing I need to fix is our local devnet script as it's broken with these changes.

PR #209, opened 2022-04-18
Remove discv5 from monorepo

### LINKED ISSUE #205: Remove `discv5` fork from monorepo

acolytec3:
Fixes #205 

This removes the `discv5` package from the monorepo.  Should not be merged until [this PR](https://github.com/ChainSafe/discv5/pull/169) is merged upstream and a new version of the discv5 dependency is released.


acolytec3:
Closing this PR.  Will replace with a new one that is built on the newly released `discv5` v0.7 that incorporates our needed changes.

PR #214, opened 2022-04-19
Initial implementation of header gossip subnetwork types

acolytec3:


packages/portalnetwork/src/headerGossipSubnetwork/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './types'
+export * from './util'
```
holgerd77, line 2: What do you think of reorganize the folder structure here a bit and do a `subnetworks` folder and then put the respective networks in folders like `state`, `history`,...?

Would find this cleaner.

holgerd77: 👍

Issue #215, opened 2022-04-20
Reorganize subnetwork directory structure


acolytec3:
Currently, all of our subnetwork (in process of rebranding to `subprotocol`) specific files are each in their own subdirectory under the main `[root]/src`.  We should move all subnetwork functionality into a subdirectory called `src/subprotocols` where each one is in its own further subdirectory (e.g. `src/subprotocols/state`, `src/subprotocol/history`, etc.).  

- [x] Move subnetwork files to a nested set of subdirectories under `src/subprotocols` #216 
- [x] Reorganize subprotocol tests in nested subprotocols directory - (oops) :grimacing: https://github.com/ethereumjs/ultralight/commit/b43ddca24b65fe23efd69e6de9004553fdd75e16
- [x] Rename files within each subprotocol from `subnetwork` to `subprotocol` as appropriate - in #225

acolytec3:
Housecleaning done

PR #216, opened 2022-04-20
Reorganize all subprotocol dirs

acolytec3:
Partially addresses #215


acolytec3:
@holgerd77 Is this what you were thinking in terms of reorganizing the subprotocol directories?

holgerd77:
Yes, that looks great, thanks! 😄 ❤️ 🙏

To be really picky: I think it would then make sense to change in the tests as well to stay consistent here. 😬

acolytec3:
Agree. Good think I didn't close the tracking issue. 😬

PR #218, opened 2022-04-22
Prototype of new rendezvous protocol

acolytec3:
Here's a prototype of the rendezvous protocol to connect to an ENR provided to us that is unreachable because of NAT/Firewall issues.  To test it out, follow these steps:

1. Start up 2 browser clients and connect them to each other.  You may have to try connecting a couple of times to connect them.
2. Start up the cli client using the below command while in the directory root:
`node packages/cli/dist/index.js --bindAddress=0.0.0.0:5500 --bootnode=enr:-IS4QFzPZ7Cc7BGYSQBlWdkPyep8XASIVlviHbi-ZzcCdvkcE382unsRq8Tb_dYQFNZFWLqhJsJljdgJ7WtWP830Gq0BgmlkgnY0gmlwhEFsKq6Jc2VjcDI1NmsxoQPjz2Y1Hsa0edvzvn6-OADS3re-FOkSiJSmBB7DVrsAXIN1ZHCCI40`

This will start the CLI client and connect to a Fluffy bootnode.  Grab the `nodeId` that the CLI client prints out to the logs for future reference.
3. Use the RPC to connect the cli client to one of your browser clients.  I use `curlie` but this can be adapted to whatever your preferred means of interacting with the rpc is:
```
curlie 127.0.0.1:8545 jsonrpc=2.0 id=1 method=portal_addBootNode params:='["enr:-[the ENR of your browser client...
```
4. Verify in the browser client logs that you see a connection established with the CLI client (this can be verified by monitoring the `discv5` and `portalnetwork` logs for successful messages exchanged with the nodeId of the cli client)
5. Now for the fun part.  Get the ENR of the cli client from the RPC - `curlie 127.0.0.1:8545 jsonrpc=2.0 id=1 method=portal_nodeEnr`
6. Go to your second browser client and try to connect to the CLI client using this ENR.  You should see errors in the logs about the messages timing out which will verify the node is unreachable.
7. Take the node ID of the CLI client and paste it into the "Target Node ID" input in the browser client dev tools. 
8. Make sure the "Select Peer" dropdown has the node ID of the rendezvous node selected (i.e. your other browser client that's connected to the CLI client).  You can verify this by looking at the Network tab in the browser client and look for a node that starts with an IP that starts with 164.92...
9. Moment of truth, click "Send Rendezvous request" and watch the logs.  If it works, you should eventually see your browser client playing portalnetwork PING/PONG with the node ID corresponding to the CLI node.


ScottyPoi:
Tested this successfully.  
Unclear to me how to incorporate this into discovery.  Would this be part of the 'firewalled' node startup?  Or is this something all nodes would do?  Or just when a node is behind a firewall?

acolytec3:
> Tested this successfully. Unclear to me how to incorporate this into discovery. Would this be part of the 'firewalled' node startup? Or is this something all nodes would do? Or just when a node is behind a firewall?

I realized I actually missed a step in the rendezvous process that's important in that the rendezvous node should send a PING to the target node before responding back to the requestor.  Otherwise, the rendezvous attempt will fail if the target node is actually unreachable from the rendezvous node (which is entirely possible if it doesn't main tight liveness checks).  

This rendezvous process should only happen when a node receives an ENR for another node that proves unreachable while it's conducting either a contentLookup or a nodeLookup.  The rendezvous process is an attempt to help the requesting node reach a target node that is is behind some sort of firewall/NAT that keeps inbound traffic out unless there is already an existing session established outbound from the target node.

PR #235, opened 2022-05-04
Web-storage

ScottyPoi:
WIP

Proof of Concept 

Use webstorage to persist node data.

Testing 2 methods:
- local storage 
- indexeddb

For storing a list of ENR's, both methods work.

List of unique ENR's will persist in the browser's indexxeddb database.

Upon reopening the page, the bootnodes sequence will also attempt to resume contact with other nodes by sending ping to each ENR.

TODO: prevent infinite growth

### DIFF
--- a/packages/browser-client/src/App.tsx
+++ b/packages/browser-client/src/App.tsx
@@ -44,6 +44,7 @@ export const mediumblue = theme.colors.blue[200]
 
 export const App = () => {
   const [portal, setPortal] = React.useState<PortalNetwork>()
+  const [IDB, setIDB] = React.useState<IDBDatabase>()
   const [peers, setPeers] = React.useState<ENR[] | undefined>([])
   const [sortedDistList, setSortedDistList] = React.useState<[number, string[]][]>([])
   const [enr, setENR] = React.useState<string>('')
@@ -92,6 +93,23 @@ export const App = () => {
   }, [portal])
 
   const init = async () => {
+    const _IDB = window.indexedDB.open('UltralightIndexedDB', 1)
+    _IDB.onupgradeneeded = () => {
+      const db = _IDB.result
+      if (!db.objectStoreNames.contains('peers')) {
+        db.createObjectStore('peers')
+      }
+      if (!db.objectStoreNames.contains('headers')) {
+        db.createObjectStore('headers')
+      }
+      if (!db.objectStoreNames.contains('blocks')) {
+        db.createObjectStore('blocks')
+      }
+    }
+    _IDB.onsuccess = () => {
+      setIDB(_IDB.result)
+      ;(window as any).IDB = _IDB.result
+    }
     if (portal?.client.isStarted()) {
       await portal.stop()
     }
--- a/packages/browser-client/src/App.tsx
+++ b/packages/browser-client/src/App.tsx
@@ -38,6 +38,7 @@ import { Capacitor } from '@capacitor/core'
 import { HamburgerIcon } from '@chakra-ui/icons'
 import Footer from './Components/Footer'
 import InfoMenu from './Components/InfoMenu'
+import { addToIndexedDB, saveToLocalStorage } from './Components/localStorage'
 // export const lightblue = '#bee3f8'
 export const lightblue = theme.colors.blue[100]
 export const mediumblue = theme.colors.blue[200]
@@ -58,7 +59,7 @@ export const App = () => {
   const { onCopy } = useClipboard(enr)
   const { onOpen } = useDisclosure()
   const disclosure = useDisclosure()
-  const toast = useToast()
+  // const toast = useToast()
   const [modalStatus, setModal] = React.useState(false)
 
   function updateAddressBook() {
@@ -153,12 +154,12 @@ export const App = () => {
     if (!errMessage) {
       errMessage = 'Node did not respond'
     }
-    toast({
-      title: errMessage,
-      status: 'error',
-      duration: 3000,
-      isClosable: true,
-    })
+    // toast({
+    //   title: errMessage,
+    //   status: 'error',
+    //   duration: 3000,
+    //   isClosable: true,
+    // })
   }
 
   async function handleFindContent(blockHash: string): Promise<Block | void> {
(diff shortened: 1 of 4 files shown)


acolytec3:
Good stuff so far.  Two things that will help with getting this right on mobile.  We should also persist the session cache if possible and also the ENR associated with the node.  Otherwise, we're just starting a new node and then establishing new sessions with each of the nodes  who's ENR is persisted to local storage.

Unfortunately for us, the session cache is a private member of a subclass of discv5.  You can access it via `(portal.client as any).sessionService.sessions` in the `browser-client` but that would then require additional hackery to instantiate just that piece.  

Can you possibly just store the entire `portal` object in local storage and then pull it out when the page reopens?  That would keep everything.  Depending on how long you go between times visiting a page, sessions could expire but that's not a big deal.

acolytec3:
I did some initial experimentation without much success but I also don't really know what I'm doing here.  I think maybe an ideal solution would persist the entire portal object to the IndexedDB/localStorage and then reopen it on page load.  Theoretically, this gives you everything, routing table, local node's ENR, session cache, DB, all in one go.  And since it looks like IndexedDB will take anything as a value, no need to worry about stringifying it first (though that shouldn't be hard since at base, we're talking about one giant JSON string).  The only thing that could be an issue longer term would be the DB, since that could get big quickly.  We could still use `level-js` to persist the DB separately and just store the portal object minus the DB property as it's own thing.  That would ensure that when the mobile app comes from background to foreground, you really are getting the current persistent node with it's associated peerId, ENR, current sessions etc.

ScottyPoi:
> Can you possibly just store the entire `portal` object in local storage and then pull it out when the page reopens? That would keep everything. Depending on how long you go between times visiting a page, sessions could expire but that's not a big deal.

My first attempt at this did not work, but I haven't tried it with indexeddb, yet.  local-storage can only store strings, and converting the portal object to JSON and back caused errors.

android's WebStorage API is slightly different but functionally should work the same.

acolytec3:
> > Can you possibly just store the entire `portal` object in local storage and then pull it out when the page reopens? That would keep everything. Depending on how long you go between times visiting a page, sessions could expire but that's not a big deal.
> 
> My first attempt at this did not work, but I haven't tried it with indexeddb, yet. local-storage can only store strings, and converting the portal object to JSON and back caused errors.
> 
> android's WebStorage API is slightly different but functionally should work the same.

Ah, right, the re-serialized portal object wouldn't be recognized as a `portalnetwork` class object.  Given that, I think maybe we can just extend your current approach to store the local node's ENR, the session cache, and the routing tables, then we just repopulate them when the page refreshes.

The flow would be something like:
- On page load, check to see if a `peerID` and `ENR` exist in local storage.  If so, use in the `portalnetwork` constructor using the peerId and ENR (so it has the same node ID)
- Repopulate the session cache using `(portal.client as any).sessionService.sessions.set()`
- Add the peers back to the routing table (as you're currently doing it more or less, though maybe without pinging each node)

ScottyPoi:
Still a WIP, but significant progress towards the goal in the browser context :)

packages/browser-client/src/App.tsx
```diff
@@ -38,28 +39,42 @@ import { Capacitor } from '@capacitor/core'
...
+  'enr:-IS4QFm4gtstCnRtOC-MST-8AFO-eUhoNyM0u1XbXNlr4wl1O_rGr6y7zOrS3SIZrPDAge_ijFZ4e2B9eZVHhmgJtg8BgmlkgnY0gmlwhM69ZOyJc2VjcDI1NmsxoQLaI-m2CDIjpwcnUf1ESspvOctJLpIrLA8AZ4zbo_1bFIN1ZHCCIyg',
+  'enr:-IS4QBE8rpfrvCZVf0RISINpHU4GM-ZmkX4y3h_WxF7YflJ-dh88a6q9_42mGVSAetfpOQqujnPE-BkDWss5qF6d45UBgmlkgnY0gmlwhJ_fCDaJc2VjcDI1NmsxoQN9rahqamBOJfj4u6yssJQJ1-EZoyAw-7HIgp1FwNUdnoN1ZHCCIyg',
+  'enr:-IS4QGeTMHteRmm-MSYniUd48OZ1M7RMUsIjnSP_TRbo-goQZAdYuqY2PyNJfDJQBz33kv16k7WB3bZnBK-O1DagvJIBgmlkgnY0gmlwhEFsKgOJc2VjcDI1NmsxoQIQXNgOCBNyoXz_7XP4Vm7pIB1Lp35d67BbC4iSlrrcJoN1ZHCCI40',
+  'enr:-IS4QOA4voX3J7-R_x8pjlaxBTpT1S_CL7ZaNjetjZ-0nnr2VaP0wEZsT2KvjA5UWc8vi9I0XvNSd1bjU0GXUjlt7J0BgmlkgnY0gmlwhEFsKgOJc2VjcDI1NmsxoQI7aL5dFuHhwbxWD-C1yWH7UPlae5wuV_3WbPylCBwPboN1ZHCCI44',
+  'enr:-IS4QFzPZ7Cc7BGYSQBlWdkPyep8XASIVlviHbi-ZzcCdvkcE382unsRq8Tb_dYQFNZFWLqhJsJljdgJ7WtWP830Gq0BgmlkgnY0gmlwhEFsKq6Jc2VjcDI1NmsxoQPjz2Y1Hsa0edvzvn6-OADS3re-FOkSiJSmBB7DVrsAXIN1ZHCCI40',
+  'enr:-IS4QHA1PJCdmESyKkQsBmMUhSkRDgwKjwTtPZYMcbMiqCb8I1Xt-Xyh9Nj0yWeIN4S3sOpP9nxI6qCCR1Nf4LjY0IABgmlkgnY0gmlwhEFsKq6Jc2VjcDI1NmsxoQLMWRNAgXVdGc0Ij9RZCPsIyrrL67eYfE9PPwqwRvmZooN1ZHCCI44',
+]
+
```
acolytec3, line 57: Let's just move these to a JSON array in it's own file and just import them and read from the array.

acolytec3 (changes requested):

acolytec3:
> Still a WIP, but significant progress towards the goal in the browser context :)

Looking great! I like the leveldb versus indexeddb.  Way simpler API.

ScottyPoi:
Go ahead and merge if it looks ok.

To clear the local storage, type LDB.clear()  into the browser console.  then next time it should start up with a new ENR, with an empty routing table and no stored_content.

acolytec3:
There's probably more work yet to be done along this path but I think we're in a good place to merge for now.

PR #237, opened 2022-05-04
Remove ENR verification check from discv5

acolytec3:
...and fork `discv5` again to make it work


packages/portalnetwork/src/client/client.ts
```diff
@@ -1130,7 +1151,8 @@ export class PortalNetwork extends (EventEmitter as { new (): PortalNetworkEvent
     }
   }
 
-  private sendPortalNetworkResponse = (
+  private sendPortalNetworkResponse = async (
+    src: INodeAddress,
```
ScottyPoi, line 1155: can we rename either `src` or `srcId`?  just being nit-picky but the similarity bugs me.
also, this refers the receiving node's info, right?  feels weird to me that what gets passed as `src` and `srcId` end up creating `dstPeerId` and `dstAddr`

packages/portalnetwork/src/client/client.ts
```diff
@@ -1148,7 +1170,9 @@ export class PortalNetwork extends (EventEmitter as { new (): PortalNetworkEvent
         return
       }
     }
-    this.client.sendTalkResp(enr, message.id, payload)
+    const dstPeerId = await enr.peerId()
+    const dstAddr = new Multiaddr(src.socketAddr.toString() + '/ipfs/' + dstPeerId.toB58String())
```
ScottyPoi, line 1174: Is this a workaround for the ENR verification check?  Or is this part of the rendezvous setup?

acolytec3: It's a workaround and I discovered it doesn't work that week with further changes I've made on the discv5 side so we have more work to do anyway.

ScottyPoi:
the message handler tests are failing.  is that just a matter of updating the test script?

acolytec3:
Probably. I hadn't gotten that far yet. This is still very much work in progress.

May 4, 2022 5:08:25 PM Scotty ***@***.***>:

packages/portalnetwork/src/client/client.ts
acolytec3: I've removed `scrId` from all of these function signatures since the id value is part of `src`.

packages/portalnetwork/src/client/client.ts
acolytec3: This should be resolved now, though it's just as hacky as this.

acolytec3:
In conjunction with the current fork of discv5, this seems to work well (and we can even connect to trin nodes again!)

We still need to test with mobile but pending any changes required in the mobile UDP transport layer, we should be close to ready to merge this if we are all on board with forking discv5 again.

acolytec3:
@ScottyPoi this ought to be ready for final review at this point.  I've fixed the bugs in the android build and all the tests are passing again.

ScottyPoi:
Looks good!

In general I think our docs could use updating.  startup is pretty simple, but still requires things like changing the ip string in the config file, starting or not starting a proxy, developer settings in the phones, and little details like that which change and can be easy to forget.

ScottyPoi (approved):

PR #246, opened 2022-05-13
Recreate-mobile

ScottyPoi:
Initial testing shows that the same browser-level API works on android, and will recreate an ENR , routing table, database.
1. Client: Added `recreateMobilePortalNetwork` modeled after `recreatePortalNetwork`
2. App: Update init() methods to look for stored mobile node
3. Unrelated Dependency Update
  -  `npm i` produced errors related to upstream dependency issues.  
  -  issues were related to `chakra-ui` releasing v2, and `react` releasign `v18`
  -  updating these did not cause any problems that i have seen.

- Not committed here are experiments with `Backgournd Mode`  which allows an app to perform tasks while running but not in the foreground.  
  - Success: 
    - App will stay active in background and continue to perform `liveness check` every 30s
  - Problems:
    - `Background Mode` requires an additional capacitor plugin (`@ionic-native/background-mode`)
    - While the app does remain active, network calls are not actually made.

### DIFF
--- a/packages/browser-client/package.json
+++ b/packages/browser-client/package.json
@@ -23,7 +23,7 @@
     "@capacitor/cli": "^3.4.3",
     "@capacitor/share": "1.1.2",
     "@types/node": "^17.0.23",
-    "@types/react": "^17.0.43",
+    "@types/react": "^18.0.9",
     "@types/react-dom": "^17.0.14",
     "assert": "^2.0.0",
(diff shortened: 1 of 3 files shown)


acolytec3 (approved):
LGTM!  I merged each of the two constructors into a single function with a flag for whether to recreate or not to simplify our code base a bit.

PR #255, opened 2022-05-24
Utp-class-reorg

ScottyPoi:
Utp class reworked as an event emitter.

No longer needs client object in the constructor, just the logger.

Emits an event for `Send` when a packet is ready to be sent, and `Stream` when all data is compiled and ready for the db

Changed `HistoryNetworkContentRequest` to `ContentRequest`, and added a `ProtocolId` attribute instead.

### DIFF
--- a/packages/portalnetwork/src/wire/utp/PortalNetworkUtp/ContentRequest.ts
+++ b/packages/portalnetwork/src/wire/utp/PortalNetworkUtp/ContentRequest.ts
@@ -4,10 +4,10 @@ import ContentWriter from '../Protocol/write/ContentWriter'
 import { sendSynPacket } from '../Packets/PacketSenders'
 import { RequestCode } from './PortalNetworkUTP'
 import { ConnectionState } from '../Socket'
+import { ProtocolId } from '../../../subprotocols'
 
-export type ContentRequest = HistoryNetworkContentRequest // , StateNetwork..., etc...
-
-export class HistoryNetworkContentRequest {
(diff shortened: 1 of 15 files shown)


packages/portalnetwork/src/wire/utp/PortalNetworkUtp/ContentRequest.ts
```diff
@@ -19,12 +19,14 @@ export class HistoryNetworkContentRequest {
...
     requestCode: RequestCode,
     contentKey: Uint8Array[],
     socket: UtpSocket[],
     socketKey: string,
     content: Uint8Array[] | undefined[]
   ) {
+    this.protocolId = protocolId
     this.sockets = socket
     //@ts-ignore
     this.contentKeys = contentKey
```
acolytec3, line 31: Is it possible to better clarify the `contentKey` vs `contentKeys` parameters that are passed into this constructor versus the properties that are being initialized?  I'm tracking why we have seeming conflicting naming patterns between them.

acolytec3 (changes requested):
Left a couple of small questions I'd like to address before we merge.

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/ContentRequest.ts
acolytec3: Nm, this change was easier than I thought.

acolytec3 (approved):
LGTM

PR #258, opened 2022-05-25
Protocol message handler

ScottyPoi:
First crack at upstreaming some of the deserialization.  Nothing new created, just deserializing in the protocol.handle() method, and sending typecast messages to the handlers, which reduces some repeat logic.

```

  public handle(message: ITalkReqMessage, src: INodeAddress) {
    const decoded = PortalWireMessageType.deserialize(message.request).value
    const id = message.id
    const messageType = message.request[0]
    const protocol = message.protocol
```

Followed by, for example: 
```
     case MessageCodes.PING:
        this.handlePing(src, id, decoded as PingMessage)
        break
```
And then handlePing() looks like:
```
  private handlePing = (src: INodeAddress, id: bigint, pingMessage: PingMessage) => {
    if (!this.routingTable.getValue(src.nodeId)) {
      // Check to see if node is already in corresponding network routing table and add if not
      this.updateRoutingTable(src.nodeId, true, pingMessage.customPayload)
    } else {
      const radius = PingPongCustomDataType.deserialize(pingMessage.customPayload).radius
      this.routingTable.updateRadius(src.nodeId, radius)
    }
    this.sendPong(src, id)
  }
  }
```

I adapted 5 tests to work with some of the changes.  I do want to check my work there to make sure it's testing properly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2010,12 +2010,6 @@
         "node": ">= 6"
       }
     },
-    "node_modules/@capacitor/cli/node_modules/tslib": {
-      "version": "2.4.0",
-      "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.4.0.tgz",
-      "integrity": "sha512-d6xOpEDfsi2CZVlPQzGeux8XMwLT9hssAsaPYExaQMuYskwb+x1x7J371tWlbBdWHroy99KnVB6qIkUbs5X3UQ==",
-      "dev": true
-    },
     "node_modules/@capacitor/core": {
       "version": "3.5.1",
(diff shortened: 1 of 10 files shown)


acolytec3:
I like where this is going.  Are you going to look at `history.ts` too?  I'm happy to pick that up if you're done at this point.

ScottyPoi:
> I like where this is going. Are you going to look at `history.ts` too? I'm happy to pick that up if you're done at this point.

Yes I can keep going with this.  Just wanted eyes on it before i dove deeper

acolytec3 (approved):
LGTM!  I added one more fix within `handleFindContent` to remove some HistoryNetwork specific stuff but this looks great.

PR #261, opened 2022-05-27
Ultralight electron app

ScottyPoi:
1. Compile the browser-client with `npm run start-browser-client`
2. Start the proxy with `npm run start-proxy`
3. from `packages/browser-client`, `npm start-electron` will start electron app.  Window shoud open automatically, and app should shutdown automatically when that window is closed.


acolytec3 (commented):
Awesome!  Looks great.

PR #273, opened 2022-06-01
Header Accumulator updates

acolytec3:
- Adds `currentHeight` method to Header Accumulator class
- Adds accumulator to History Protocol
- Update accumulator when a new header is found that is the next header in the chain

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -510,7 +510,7 @@
     "node_modules/@babel/highlight/node_modules/escape-string-regexp": {
       "version": "1.0.5",
       "resolved": "https://registry.npmjs.org/escape-string-regexp/-/escape-string-regexp-1.0.5.tgz",
-      "integrity": "sha1-G2HAViGQqN/2rjuyzwIAyhMLhtQ=",
+      "integrity": "sha512-vbRorB5FUQWvla16U8R/qgaFIya2qGzwDrNmCZuYKrbdSUMG6I1ZCGQRefkRVhuOkIGVne7BQ35DSfo1qvJqFg==",
       "engines": {
         "node": ">=0.8.0"
       }
@@ -518,7 +518,7 @@
     "node_modules/@babel/highlight/node_modules/has-flag": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/has-flag/-/has-flag-3.0.0.tgz",
-      "integrity": "sha1-tdRU3CGZriJWmfNGfloH87lVuv0=",
+      "integrity": "sha512-sKJf1+ceQBr4SMkvQnBDNDtf4TXpVhVGateu0t918bl30FnbE2m4vNLX+VWe/dpjlb+HugGYzW7uQXH98HPEYw==",
       "engines": {
         "node": ">=4"
       }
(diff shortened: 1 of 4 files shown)


acolytec3:
The idea is that every node maintains an accumulator which contains all of the header hashes from the entire chain (up to the current chain height that the node is aware of).  Using the magic of SSZ merkle-ization, a node can receive any header with a merkle proof derived from the header accumulator and validate whether the header is indeed part of what that node knows to be the canonical chain.  In other words, if my current accumulator corresponds to the chain at height 10000, if I request a block header for any block up to 10000, if the node that sends me that header + a merkle proof of it's inclusion in the accumulator, I can be confident that the block they sent me is indeed the actual block from the canonical chain as opposed to an uncle that is mined at the same height.  Does that make sense?

acolytec3:
Closing in favor of #286

holgerd77:
> The idea is that every node maintains an accumulator which contains all of the header hashes from the entire chain (up to the current chain height that the node is aware of). Using the magic of SSZ merkle-ization, a node can receive any header with a merkle proof derived from the header accumulator and validate whether the header is indeed part of what that node knows to be the canonical chain. In other words, if my current accumulator corresponds to the chain at height 10000, if I request a block header for any block up to 10000, if the node that sends me that header + a merkle proof of it's inclusion in the accumulator, I can be confident that the block they sent me is indeed the actual block from the canonical chain as opposed to an uncle that is mined at the same height. Does that make sense?

Ah, that makes a lot of sense, thanks! Why is this not part of the general protocol, adding an accumulator every 10.000 blocks or so? 🤔 (or do they have something like this in Eth 2?)

acolytec3:
> Ah, that makes a lot of sense, thanks! Why is this not part of the general protocol, adding an accumulator every 10.000 blocks or so? thinking (or do they have something like this in Eth 2?)

It does exist in the consensus layer, though I'd have to dig to figure out where it was.  @g11tech might be able to point you in the right direction.  

I know there's been discussion around the Portal Network world about possibly trying to get it into an EIP to include in the Execution Layer at some point but I'm not aware that one has started at this point.

g11tech:
> > Ah, that makes a lot of sense, thanks! Why is this not part of the general protocol, adding an accumulator every 10.000 blocks or so? thinking (or do they have something like this in Eth 2?)
> 
> It does exist in the consensus layer, though I'd have to dig to figure out where it was. @g11tech might be able to point you in the right direction.
> 
> I know there's been discussion around the Portal Network world about possibly trying to get it into an EIP to include in the Execution Layer at some point but I'm not aware that one has started at this point.

we have block roots and state roots in beacon state
```typescript
 export const BeaconState = new ContainerType(
  {
    genesisTime: UintNum64,
    genesisValidatorsRoot: Root,
    slot: primitiveSsz.Slot,
    fork: phase0Ssz.Fork,
    // History
    latestBlockHeader: phase0Ssz.BeaconBlockHeader,
    blockRoots: HistoricalBlockRoots,
    stateRoots: HistoricalStateRoots,
    historicalRoots: new ListCompositeType(Root, HISTORICAL_ROOTS_LIMIT),
    // Eth1
    ```

PR #277, opened 2022-06-02
client: set bucketRefresh interval on start

ScottyPoi:
bucket refresh was disabled in recent refactor

This PR brings back `setInterval()` for bucket refresh as part of `client.start()`

`protocol.bucketRefresh()` was set to `private`, and was unaccessable from client.  making this method `public` solved the issue.  

If we prefer this method to remain private we will have to move the setInterval() call into `protocol` itself


acolytec3 (approved):
LGTM

PR #285, opened 2022-06-06
Protocol test reorganization

acolytec3:
- Moves Portal Network wire message tests from `client.spec.ts` to corresponding test files at correct subdirectory level within `subprotocols` 
- Adds new `handleFindNodes` tests to validate various `FOUNDNODES` message response scenarios
- Fixes logic bug in how `FOUNDNODES` message is constructed


acolytec3:
I'm going to go ahead and merge this as I'm reasonably confident I've fixed at least one logic bug that was limiting our node responses.

PR #290, opened 2022-06-09
Content Gossip

ScottyPoi:
Inserted content gossip method at the end of `addContentToHistory()`

Removed the `radius` check because it was returning undefined, causing no gossip to happen.  Receiving node can reject OFFER if it is not within their radius.

I also made changes to the logger while i was working on this, and  committed the changes I made.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -116,7 +116,7 @@ const main = async () => {
     dataDir: args.datadir,
   })
   portal.discv5.enableLogs()
-  portal.enableLog('*ultralight*, *portalnetwork*, *uTP*, *discv5*')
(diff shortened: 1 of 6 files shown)


acolytec3:
Is there a definitive need for this pattern `this.logger.extend(`NODES`)(...`?  I'm okay with it but I had to hack aroudnd for a while and my solution requires some hideous monkey patching to make the tests pass with it.  It's probably partially my lack of skill with `testdouble` but it's going to make adding more tests occasionally challenging unless we figure out how to make the testdouble I created for the logger work right.

ScottyPoi:
No definitive need for any of it.  I was working on the content gossip and the logs had become too noisy to be helpful.   I didn't set anything up yet but I looked into Loki for getting metrics on logs.

ScottyPoi:
What is testdouble doing?

PR #291, opened 2022-06-09
Fix handleFindNodes logic and tests

acolytec3:
Fixes the too clever by half logic in `handleFindNodes` as well as removes test hackery that was trying to go around my elbow to scratch my nose.


acolytec3:
@ScottyPoi handleFindNodes logic should be fixed now

PR #293, opened 2022-06-10
protocol: more aggressive bucketRefresh

ScottyPoi:
-  During bucketRefresh, we now perform a node lookup for all buckets which are not full.  
-  During addBootnode, we now send separate findNodes messages for each bucket, as we are hopefully receiving lots of ENR's during that phase.
- Extraneous bucketRefresh logging was removed
- Duplicate Discv5 logging was removed from CLI

- Added portal_nodeLookup to rpc methods
- Node lookup was fixed to properly send PING on successful lookup

- Added some defaults to devnet.ts yargs
  - path/to/pks.txt
  - numNodes: 16
- Committed some helper scripts for test setups


acolytec3:
All the tests are passing so I'm inclined to merge this one.  I've done some local testing though still need to work on some tuning to see if I can observe the stepdown logic in action but the numbers I proposed seem like maybe a good starting point for a large network with lots of nodes in it.  Not sure how full we expect peoples' routing tables to be.

acolytec3 (approved):
Approving as I'm happy with this for now.  Go ahead and merge if you're good with the changes.

PR #294, opened 2022-06-10
Browser-getBlockByNumber

ScottyPoi:
Exposes portal_getBlockByNumber in the browser client.

Adds an input and button to app screen for search by number.  

Refactored how `block` and `setBlock` are transferred.  useContext, useState, and useMemo combine to pass down both the prop and the setter without the need to pass props and setter functions through each component in the chain.


acolytec3 (approved):
I'm good with this as far as it goes.  getBlockByNumber won't work at this point since the browser client doesn't have its canonical index locally so it's not really testable at this point.  I'm going to work on code to transfer the accumulator over the wire in the morning and we can maybe bake in some updates to the canonical indices network code so it rebuilds the block index from an accumulator snapshot if received and then this can probably work okay.

PR #297, opened 2022-06-21
Remove `syntheticdefault` and `esModuleInterop`

acolytec3:
Removes `syntheticDefault` and `esModuleInterop` from our tsconfig in preparation for putting together and ESM build


ScottyPoi (approved):
Wow, thanks.  I find the whole thing to be *ugly*  compared to how it looks with the options on, but I do enjoy taking off the training wheels.  

The tests pass so I'm happy

PR #312, opened 2022-07-14
getBlockByNumber by accumulator

acolytec3:
- Removes Canonical Indices subprotocol and references to block index
- Revise `history.getBlockByNumber` to use header accumulator to retrieve block hash from number


ScottyPoi (approved):
LGTM.

PR #313, opened 2022-07-15
New-utp-tests

ScottyPoi:
Updates outdated tests, and creates new tests for current uTP structure.
Also changes uTP packet creators, `createSynPacket`, etc. to `Packet.create(packetType)

Adjusts some utp code to enable more testing.

Fixes bugs exposed by testing.

Current tests cover:
1. uTP Packet ecoding/decoding
2. VarInt prefix compression / decompression
3. uTP `ContentWriter` and `ContentReader` (chunking and compiling)
4. run `cli/scripts/sendOffer.ts` with 2 cli nodes running to  test OFFER/ACCEPT


acolytec3 (commented):
Love the new modularization and tests!  Feels more readable now.

packages/portalnetwork/src/wire/utp/Protocol/write/ContentWriter.ts
```diff
@@ -62,18 +63,18 @@ export default class ContentWriter {
...
-  chunk(content: Uint8Array, size: number): Record<number, Uint8Array> {
-    let arrayMod = content
+  chunk(): Record<number, Uint8Array> {
+    let arrayMod = this.content
     this.logger(`Preparing`)
-    this.logger(content)
-    this.logger(`For transfer as ${size} byte chunks.`)
-    const full = Math.floor(content.length / size)
-    const partial = content.length % size > 0 ? 1 : 0
+    this.logger(this.content)
```
acolytec3, line 69: Is there a reason for us to dump the entire bytestring to the logs? Maybe we could log just the first 20 bytes or something if it has some use but otherwise just feels overly verbose.

ScottyPoi: No, not at all.  Just a relic from me working on it. I just like seeing each thing at every stage while I build.  I'm starting to understand how writing tests as I go accomplishes the same thing and more, but *i'm just a baby*

acolytec3 (approved):
This looks merge-worthy to me.  Anything else to add before we do so?

PR #314, opened 2022-07-18
Add tauri build

acolytec3:
I'm not sure if we want to include this or not but this adds the steps to build a desktop app using [tauri](https://tauri.app/).  There was absolutely zero configuration necessary on the react side to get this to build and run with the dev server.  

And (I think it's a lot lighter than electron in the final build step).  Here's the [guide to the necessary build dependencies for Linux](https://tauri.app/v1/guides/getting-started/prerequisites#setting-up-linux)

Use `ctrl + shift + i` to get the dev console open.  

Also, you need to run the proxy as I haven't written a transport layer to use UDP with this app (though presumably it would be pretty simple).


ScottyPoi:
LGTM -

Should we remove the electron app?

acolytec3:
> LGTM -
> 
> Should we remove the electron app?

No, electron is a more popular build platform at this point and demonstrating that it works on multiple platforms is just another good proof of concept.  Once I get around to the new transport layers, we'll have more ready made options for teams to build with.

ScottyPoi (approved):

holgerd77:
Pretty cool ❤️ 🎉 - that's really interesting - can someone post a screenshot here?

acolytec3:
> Pretty cool heart tada - that's really interesting - can someone post a screenshot here?

Here you go.  Basically the same as the existing mobile app but runs in a pure webview instead of chromium

image

PR #318, opened 2022-07-25
Add new history ssz types for BlockBody and Receipts

acolytec3:


ScottyPoi (approved):
Nice!  Very organized and neat code all around.

PR #320, opened 2022-07-28
Rpc-refactor

ScottyPoi:
Adds `Receipt Manager` (borrowed from EthJS) to history protocol.  This is a placeholder while we figure out actual receipt management.

Follows the model of EthereumJS Client's RPC module.  Current RPC methods are separated into `portal` and `eth` modules.  

Functionally it should feel exactly the same.  Devnet script and the RPC calls set up in current test scripts should all still function without change.


acolytec3 (changes requested):
I will review more throughout the day but have left some small feedback.

acolytec3 (approved):
On second thought, LGTM.  Not sure about the receipts manager and how we want to handle that bit so will leave in for now (though I'm wondering if we'll end up removing it later or else move it out of `portalnetwork`).

PR #329, opened 2022-08-08
fix: set correct max within the `randUint16` function and remove unused `randUint32` function

### LINKED ISSUE #274: uTP Out-of-Range Error
If random seqNr is generated too close to MAX_UINT16, uTP throws RangeError when seqNr incrementing goes over MAX.

```
RangeError [ERR_OUT_OF_RANGE]: The value of "value" is out of range. It must be >= 0 and <= 65535. Received 65536
    at new NodeError (node:internal/errors:371:5)
    at checkInt (node:internal/buffer:72:11)
    at writeU_Int16BE (node:internal/buffer:832:3)
    at Buffer.writeUInt16BE (node:internal/buffer:840:10)
    at packetToBuffer (/home/scottypoi/Documents/github/ultralight/packages/portalnetwork/dist/wire/utp/Packets/Packet.js:21:12)
    at Packet.encodePacket (/home/scottypoi/Documents/github/ultralight/packages/portalnetwork/dist/wire/utp/Packets/Packet.js:55:24)
```

Need a guard against seqNr's that climb past MAX_UINT16

Possible Solutions:
    - Flat ceiling for randUint16()  (Like MAX - 1000)

l2ig:
Fixes #274

### DIFF
--- a/packages/portalnetwork/src/wire/utp/Utils/math.ts
+++ b/packages/portalnetwork/src/wire/utp/Utils/math.ts
@@ -15,9 +15,6 @@ export function Bytes32TimeStamp(): number {
 export function randUint16(): Uint16 {
   return Math.floor(Math.random() * (2 ** 16 - 1))
 }
-export function randUint32(): Uint16 {
-  return Math.floor(Math.random() * 2 ** 32)
-}
 
 export function bitLength(n: number): number {
   const bitstring = n.toString(2)
--- a/packages/portalnetwork/src/wire/utp/Utils/math.ts
+++ b/packages/portalnetwork/src/wire/utp/Utils/math.ts
@@ -13,7 +13,7 @@ export function Bytes32TimeStamp(): number {
 }
 
 export function randUint16(): Uint16 {
-  return Math.floor(Math.random() * 2 ** 16)
+  return Math.floor(Math.random() * (2 ** 16 - 1))
 }
 export function randUint32(): Uint16 {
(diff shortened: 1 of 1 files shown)


packages/portalnetwork/src/wire/utp/Utils/math.ts
```diff
@@ -13,10 +13,7 @@ export function Bytes32TimeStamp(): number {
 }
 
 export function randUint16(): Uint16 {
-  return Math.floor(Math.random() * 2 ** 16)
-}
-export function randUint32(): Uint16 {
-  return Math.floor(Math.random() * 2 ** 32)
+  return Math.floor(Math.random() * (2 ** 16 - 1))
```
acolytec3, line 16: @ScottyPoi I'm thinking we need something lower than just 2**16 -1 as our max, right?.  Isn't the issue that we could pick a sequence number for uTP too close to the maxium (2**16) and overflow when sending a long stream of packets (say a header accumulator that's 2mb and requires 2000 packets)?

acolytec3:
Thanks for the PR!  I've left a comment for @ScottyPoi on this so will wait his feedback before we consider merging.

packages/portalnetwork/src/wire/utp/Utils/math.ts
ScottyPoi: This is addressed in #331
I just used `2**15` as the max

PR #332, opened 2022-08-10
Move handle find content

acolytec3:
We had some history protocol specific stuff seep into `handleFindContent` so I think it makes sense to make this a subprotocol specific method so we can do more specific handling within each subprotocol.  I'm motivated mainly by the fact that we want to start validating headers against our accumulator and this should be done in handleFindContent if possible.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9751,12 +9751,6 @@
         "node": ">=0.10.0"
       }
     },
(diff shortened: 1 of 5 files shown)


acolytec3:
Nevermind, having a "not had enough coffee this morning" moment.

PR #335, opened 2022-08-12
Add proxy address reflection back to browser-client

acolytec3:
Adds public multiaddr reflection back to browser client


acolytec3:
Note, this only works when the `BINDADDRESS` environmental variable is set the browser-client (i.e. when `npm run start-local` is executed.

PR #337, opened 2022-08-12
Fix missing SYN-ACK message

acolytec3:
Add back SYN-ACK handler in FINDCONTENT_WRITE


packages/portalnetwork/src/wire/utp/PortalNetworkUtp/PortalNetworkUTP.ts
```diff
@@ -284,7 +284,8 @@ export class PortalNetworkUTP extends BasicUtp {
         this.logger(`SYN received to initiate stream for FINDCONTENT request`)
         request.socket.ackNr = packet.header.seqNr
         request.socket.seqNr = randUint16()
-        writer = await this.createNewWriter(request.socket, request.socket.seqNr)
+        await this.sendSynAckPacket(request.socket)
+        writer = await this.createNewWriter(request.socket, request.socket.seqNr++)
```
ScottyPoi, line 288: Also adding the seqNr++ to the writer creation.  Otherwise data packet 1 will have same seqNr as syn-ack

ScottyPoi (approved):

PR #343, opened 2022-08-16
Simplify `contentLookup`

acolytec3:
Simplifies return type of `contentLookup` so it only returns the desired content or else `undefined`.  
Also cleans up the eventListener associated with each content lookup when the content lookup resolves.  
Lastly, allows for simplification of `getBlockByHash` and `getBlockByNumber`.

### DIFF
--- a/packages/portalnetwork/src/subprotocols/contentLookup.ts
+++ b/packages/portalnetwork/src/subprotocols/contentLookup.ts
@@ -2,6 +2,7 @@ import { ENR, distance, NodeId, EntryStatus } from '@chainsafe/discv5'
 import { fromHexString, toHexString } from '@chainsafe/ssz'
 import { Debugger } from 'debug'
 import { serializedContentKeyToContentId, shortId } from '../util/index.js'
+import { HistoryNetworkContentTypes } from './history/types.js'
 import { BaseProtocol } from './protocol.js'
 
 type lookupPeer = {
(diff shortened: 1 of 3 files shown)


packages/portalnetwork/src/subprotocols/contentLookup.ts
```diff
@@ -79,8 +81,19 @@ export class ContentLookup {
           this.logger(`received uTP connection ID from ${shortId(nearestPeer!.nodeId)}`)
           finished = true
           nearestPeer.hasContent = true
-          return res.value
+          return new Promise((resolve) => {
+            const utpDecoder = (
```
ScottyPoi, line 85: Is it an technical choice or style to define this in the return, instead of defining `utpDecoder` on its own?

Maybe i'm just being a n00b here, but this kind of syntax is a little confusing to me.

ScottyPoi (approved):

ScottyPoi:
Nice, the app is displaying headers now that come in without bodies.

packages/portalnetwork/src/subprotocols/contentLookup.ts
acolytec3: I thought about defining it as a general utility method of sorts but this is a pattern I learned for cleaning up event listeners after their intended purpose is complete. You have to pass the exact function you put on the listener to the `removeListener` or it won't work and we don't want to remove all Content Added listeners. There are probably better ways to do it but I haven't found one yet.

PR #344, opened 2022-08-17
Peer Tools in Network Tab

ScottyPoi:
Updates the `netowrk` tab in browser client to include tools to interact with individual peers.

Address book view is unchanged, but has become a select menu to choose a peer to interact with.

PeerTools will appear over (small screen) or beside (large screen) the address book.

PeerTools expands on DevTools with an improved interface to test Ping/Pong, FindContent, and Offer.

Also extracts the `add peer` function to the main page of the app.


acolytec3:
Excellent!  I removed the duplicate stuff from the devtools overlay but otherwise looks great!

PR #345, opened 2022-08-18
DX-for-the-win

ScottyPoi:
Implements `useContext` to reduce the amount of props passed between components.

Moves `createContext` to separate file.

Creates `Header` component instead of building in `App.tsx`.

Moves `copyENR` and `Upload from file` buttons to Header.

Removes hamburger menu for DevTools*
  * TODO: relocate  StartNode `connect to proxy` button

Consolidates `HistoryNetwork` component into `Layout`, removes disabled subprotocol tab menu.

Creates component for getBlockByHash.

Renames ambiguous function `handleClick` --> `connectToPeer`


acolytec3:
While I will look at the code in more detail tomorrow, if I search by block number for something that's higher than whats in our current accumulator, the react app breaks.  The logs show it's failing to find a block, but unfortunately the logs aren't clear as to which component is breaking.  Something inside `Layout` but that's not very precise.  Will do some more digging in the morning.

acolytec3 (approved):
Looks great!

PR #346, opened 2022-08-19
Validate header matches hash in findContent responses

acolytec3:
- Verifies that constructed header matches sought hash in `addContentToHistory`
- Adds `Offer Snapshot` button to browser client
- Fixes `peer` onclick in routing table view


ScottyPoi (approved):
LGTM

PR #351, opened 2022-08-26
Fixing typos

omahs:


acolytec3:
Thanks for this contribution but we're not soliciting small readme updates at this point.  We'll do clean-up of our readmes down the line and would be open to more comprehensive contributions then.

PR #352, opened 2022-08-26
Header proof buttons

ScottyPoi:
Adds buttons to browser-client layout

*  `Header Proof By Blockhash` 
    * Displays a success or failure with toast
* `Validate accumulator`
    * Treats history.accumulator as a shapshot, and runs validation protocol
    * Displays a success or failure with toast


acolytec3:
Approved

PR #356, opened 2022-08-30
History-Protocol-Management

ScottyPoi:
* **Reduces `portalnetwork/srcc/subprotocols/history.ts` from **659** lines to a much more digestible **205** by outsourcing to different managers.**

1. AccumulatorManager
    - Manages needs relating to HeaderAccumulator, EpochAccumulator, and HeaderProofs
2. ETH
    - Manages `eth_` namespace methods (getBlockBy...)
3. GossipManager
    - Manages content gossip
4. ContentManager
    -  Manages logic previously done in `history.addContentToHistory`

* Also generally reduces lines of code throughout the repo through updates to `index.ts` files and consolidation of `imports`.


ScottyPoi:
## Gossip Manager

Absorbs the gossipQueue and `gossipHistoryNetworkContent` methods, but also adds more sophisticated logic to the content gossip mechanism.

Instead of one master gossip queue, an individual queue is managed gossip to that peer is only offered when there are a number of items in that queue.

This reduces the number of OFFER/ACCEPT handshakes that need to occur, without reducing the actual flow of gossiped content.

acolytec3:
I like the general approach.  One thing though brought over from the other monorepo is that we should avoid the use of getters and setters. I think I roll back the `get header` and `get currentHeight` properties on the accumulator manager.  

Separately, I cleaned up some more unused imports and consolidated the gossip manager down a little bit and then added some docs.

acolytec3:
Some of the integration tests are failing so need to fix that but this removes the getters and replaces them with general functions.

packages/portalnetwork/src/subprotocols/history/headerAccumulator.ts
```diff
@@ -103,26 +103,25 @@ export class AccumulatorManager {
...
-  public get header() {
-    return this.headerAccumulator
-  }
-
   public updateAccumulator(newHeader: BlockHeader) {
     this.headerAccumulator.updateAccumulator(newHeader)
   }
 
-  public get currentHeight() {
+  public currentHeight() {
```
ScottyPoi: :+1: 

I prefer this approach, anyway.  Whoever built the headerAccumulator must have left some of their `get`s around...

packages/portalnetwork/src/subprotocols/history/history.ts
```diff
@@ -47,7 +47,12 @@ export class HistoryProtocol extends BaseProtocol {
     this.contentManager = new ContentManager(this, nodeRadius ?? 4n)
   }
 
-  public _handleFindContent = async (decodedContentMessage: FindContentMessage) => {
+  /**
+   *
+   * @param decodedContentMessage content key to be found
+   * @returns content if available locally
+   */
+  public findContentLocally = async (decodedContentMessage: FindContentMessage) => {
```
ScottyPoi, line 55: I don't *love* the name - but i can learn to accept it

packages/portalnetwork/src/subprotocols/history/headerAccumulator.ts
acolytec3: Probably, never can trust that guy.

ScottyPoi:
Changes look good.

acolytec3 (approved):

PR #357, opened 2022-09-07
Implement useReducer and useContext in react app

ScottyPoi:
Unifies state into global useReducer and useContext.

 -  useReducer consolidates most of the state and state change functions into one global `state`, and a set of state change functions called from `dispatch`.
 - This, combined with useContext, allows the global state and dispatch to be consumed by all downstream components without the need to pass props.
 - The Developer tools (`PeerButtons`) has its own state and reducer and useContext, which is consumed only by itself and downstream buttons
 - Some components still maintain their own state, where the values only affect the component itself.

* Also adds tabs to view receipts to block explorer.  
* And a tab to view individual receipts per transaction.
* Other minor layout and housekeeping changes


ScottyPoi:
@acolytec3 

I feel about 90% confident that all the components work same as before.  The app could use a thorough test drive before merging this, though, in case I missed anything.  This doesn't *add* anything new, just changes how state is handled

packages/browser-client/src/Components/PeerButtons.tsx
```diff
@@ -24,360 +24,418 @@ import {
...
+  SETBLOCKHASH = 'SETBLOCKHASH',
+  SETPEERIDX = 'SETPEERIDX',
+}
+
+interface AppStateAction {
+  type: PeerButtonsStateChange
+  payload?: any
+}
+
+const reducer = (state: PeerButtonsState, action: AppStateAction) => {
```
acolytec3, line 60: Can we put this reducer in a separate utility file and keep all this state management logic out of the front-end?  I don't want to overengineer this but since we're baking all this stuff in, it would be nice to have the reducers and state management logic all in their own subdirectory.

acolytec3 (changes requested):
This looks great overall but two general things I'd like to change.  
1) Let's remove any commented code you don't need
2) Lets put the reducers and utility functions in their own subdirectory so under `src/reducers` or `src/context` or something and pull all the state management logic and other helper methods there so we have a proper separation of concerns between front-end and "business" logic.  That'll make it easier when I implement the `provider` interface in `portalnetwork` to just touch things in one place.

acolytec3 (approved):
Going to go ahead and merge.  We should probably do some additional refactoring along the way to pull the reducers/context into a separate subdirectory.  Will add an issue on that.

Issue #358, opened 2022-09-13
Isolate business logic and UI components in the browser client


acolytec3:
#357 updated our browser client to use React useReducer/useContext for most stuff.  There's a separate reducer inside of the `PeerButtons` component that we should also pull into a separate utility file and then work on making our app follow general webdev best practice of separating business logic and UI stuff for easier reasoning about the code.

acolytec3:
Fixed by #361

PR #361, opened 2022-09-19
browser-client: move PeerReducer into separate utility file

ScottyPoi:
for issue #358
continuing PR #357

Clean up for useReducer implementation in the `PeerButtons` component.

Creates new utility files:
* peerReducer.ts
  * Handles the state / dispatch logic
* peerActions.ts
  * Handles the portal network calls triggered by clicking buttons

### DIFF
--- a/packages/browser-client/src/Components/PeerButtons.tsx
+++ b/packages/browser-client/src/Components/PeerButtons.tsx
@@ -26,61 +26,11 @@ import {
 } from 'portalnetwork'
 import React, { useContext, useEffect, useReducer } from 'react'
 import { AppContext, AppContextType, StateChange } from '../globalReducer'
-
-interface PeerButtonsState {
-  epoch: number
-  offer: Uint8Array[]
-  ping: [string, string]
-  distance: string
-  blockHash: string
-}
-
-const peerButtonsInitialState = {
-  epoch: 0,
-  offer: [],
-  ping: ['blue.100', 'PING'] as [string, string],
-  distance: '',
-  blockHash: '',
-}
-
-enum PeerButtonsStateChange {
-  SETEPOCH = 'SETEPOCH',
-  ADDTOOFFER = 'ADDTOOFFER',
-  PING = 'PING',
-  SETDISTANCE = 'SETDISTANCE',
-  SETBLOCKHASH = 'SETBLOCKHASH',
-  SETPEERIDX = 'SETPEERIDX',
-}
-
-interface AppStateAction {
-  type: PeerButtonsStateChange
-  payload?: any
-}
-
-const reducer = (state: PeerButtonsState, action: AppStateAction) => {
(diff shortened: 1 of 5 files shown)


packages/browser-client/src/Components/PeerButtons.tsx
```diff
@@ -421,16 +238,16 @@ export default function PeerButtons() {
                 width={'33%'}
                 title="Add content to offer"
                 onClick={() => {
-                  addToOffer('block')
+                  peerActions.addToOffer('block')
```
acolytec3, line 241: Does it make sense to have this "Offer Block" since the idea of these buttons is to allow direct control of the wire methods of the protocol?  We don't have a concept of offering a "block" since a block isn't a discrete piece of content in the portal network.  If we remove Find/Offer Block, we can greatly simplify the logic of the PeerReducer and I dunno if it really gains us anything.

packages/browser-client/src/Components/PeerButtons.tsx
```diff
@@ -403,7 +220,7 @@ export default function PeerButtons() {
                 width={'33%'}
                 title="Add content to offer"
                 onClick={() => {
-                  addToOffer('header')
+                  peerActions.addToOffer('header')
```
acolytec3, line 223: The hover state `title` element of the "Find Header"/"Find Body" buttons doesn't match.  It says `add content to offer` which seems incorrect.

packages/browser-client/src/peerActions.ts
```diff
@@ -0,0 +1,178 @@
...
+  state: PeerState
+  dispatch: PeerDispatch
+  historyProtocol: HistoryProtocol
+  constructor(peerContext: PeerContextType, protocol: HistoryProtocol) {
+    this.state = peerContext.peerState
+    this.dispatch = peerContext.peerDispatch
+    this.historyProtocol = protocol
+  }
+
+  addToOffer = (type: string): void => {
```
acolytec3, line 21: If we remove the "Offer Block" buttons, we can simplify this whole method to just accept a `HistoryNetworkContentTypes` "type" parameter and then put that value in the `selector` field of the `HistoryNetworkContentKeyUnionType...` serializer and remove the whole switch statement.

acolytec3 (changes requested):
I think we should consider removing the whole concept of "Find/Offer" block from the `Peer Buttons` menu.  We already have the abstract concept of finding a block represented by the Find Block By Hash/Number buttons.

packages/browser-client/src/Components/PeerButtons.tsx
ScottyPoi: Yeah I suppose it does diverge from the point of the peer controls, since it does a bunch of extra logic and isn't pure Wire Protocol.

acolytec3 (approved):
Lgtm

PR #364, opened 2022-09-19
Integrate Ultralight provider into Browser-client

acolytec3:
Integrated the initial pass of the provider with the browser client


ScottyPoi:
LGTM -- The react app is buggy in a couple ways, but can be fixed (or maybe already has) in another PR.  The Provider stuff looks good, though!

ScottyPoi (approved):

Issue #374, opened 2022-10-04
Remove "master accumulator" 


ScottyPoi:
*part of meta-issue #372 *
## Remove the "master accumulator" as a network supported data type.

Master Accumulator for premerge blocks will be hardcoded, and will not be gossiped around the network.  

- [x] Remove master accumulator from history network content types
- [x] Remove tests for building and sharing master accumulator
- [x] Initialize History Network protocol with verified pre-built accumulator

Issue #373, opened 2022-10-04
Redefine content keys


ScottyPoi:
*part of meta-issue #372 *

- [x]  #375 Remove Chain Id from contentKeys
- [x]  Remove the global definition of how content_key is derived in favor of explicit encoding that doesn't implicitly rely on SSZ encoding semantics. (Implementations should still be able to implement things as an SSZ Union if they would like).
- [x] Add a content_key definition to every data type supported by the network.

Issue #365, opened 2022-09-19
Remove chain id from data structures


ScottyPoi:
*part of meta-issue #372 *

### Remove ChainId

Following discussion around https://github.com/ethereum/portal-network-specs/issues/163

PR #378, opened 2022-10-09
History-network-updates

ScottyPoi:
Updates History Protocol with changes to portalnetwork specs.

1.  Removes `chainId`  from contentKey data structures 
 - #365  
 - #375
2.  Removes `MasterAccumulator` from network content types / Adds pre-built master accumulator
- #374 
- #376 
    - TODO: Build or borrow a fully build pre-merge accumulator
3.  Redefines content keys
- #373 
- #377


ScottyPoi:
Not 100% confident that I implemented #373 correctly.

@acolytec3  can you read that part of the spec changes and tell me if you interpret it the same as I do?

packages/browser-client/src/peerActions.ts
```diff
@@ -81,15 +81,15 @@ export class PeerActions {
       const block = reassembleBlock(header!.value as Uint8Array, undefined)
       return block //
     } else if (type === 'body') {
-      const headerKey = HistoryNetworkContentKeyUnionType.serialize({
+      const headerKey = HistoryNetworkContentKeyType.serialize({
         selector: 0,
         value: {
           chainId: 1,
```
acolytec3, line 87: Do we need to remove chainID here as well?  I'm assuming we never use chainId anywhere, right?

packages/portalnetwork/src/subprotocols/history/headerAccumulator.ts
```diff
@@ -77,6 +82,27 @@ export class HeaderAccumulator {
     }
     return this._currentEpoch.push(headerRecord) - 1
   }
+  /**
+   * Adds a new block header to the `currentEpoch` in the header accumulator
+   */
+  public updateAccumulatorFromEthers = (newHeader: { hash: string; difficulty: number }) => {
```
acolytec3, line 88: Do we need this anymore?  Seems like it's of limited value since we have a hardcoded master accumulator.  Can't we rip out all of this logic for updating the master accumulator?

acolytec3 (commented):
Left a couple of questions for us to think through before merging.

packages/browser-client/src/peerActions.ts
ScottyPoi: Ah, this is outdated.  content keys are not ssz anymore.

packages/portalnetwork/src/subprotocols/history/headerAccumulator.ts
ScottyPoi: we can.  What seemed like minor changes had already created a monster PR.  But yes, ripping all of that out is fine now that we have the pre-merge accumulator set in stone

packages/browser-client/src/peerActions.ts
ScottyPoi: Fixed -- Same thing now looks like
```
const headerContentId = getHistoryNetworkContentId(BlockHeader, this.state.blockHash)

packages/portalnetwork/src/subprotocols/history/headerAccumulator.ts
ScottyPoi: Went ahead and axed all of the updating and snapshotting stuff.

acolytec3 (approved):
LGTM

PR #381, opened 2022-10-28
Eviction 3 strikes rule

ScottyPoi:
#### Node eviction logic

- [x] Fixes issue with Ultralight nodes sending PING requests long after supposedly "evicting" a peer.
- [x] Fixes issue with Ultralight nodes trying to evict the same peers over and over

1. Using `routingTable.remove(enr)` instead of `routingTable.removeById(nodeId)` seems to work better
2. Peers are now given 3 chances to respond to PING messages before eviction
3. A successful PING/PONG will reset the strikes. 
4. A liveness check will PING a 20% sample at a regular interval, and log the results.

- Implements a 3 strikes rule for peers failing to respond to PING messages.
- Edits liveness check to use 20% sample of peers
- Enables liveness check in bucketRefresh
- Reduces bucketRefresh to 4 buckets per refresh.
- Reduces logging output for NODES/FINDNODES when result is 0 nodes.
- Adds logger to routing table.
- Sets routing table logger in protocol constructor
- Adds mapping of <nodeId, strikes> to routing table
- Adds `strike` and `clearStrikes` methods to routing table.
- Edits `evict` method to evict using ENR
- Implements 3 strikes rule for evicting Peers
- Adds logic to `sendPing` to add / clear strikes

### DIFF
--- a/packages/portalnetwork/src/client/routingTable.ts
+++ b/packages/portalnetwork/src/client/routingTable.ts
@@ -1,12 +1,33 @@
 import { KademliaRoutingTable, NodeId } from '@chainsafe/discv5'
-
+import { Debugger } from 'debug'
 export class PortalNetworkRoutingTable extends KademliaRoutingTable {
+  public logger?: Debugger
   private radiusMap: Map<NodeId, bigint>
   private gossipMap: Map<NodeId, Set<string>>
+  private strikes: Map<NodeId, number>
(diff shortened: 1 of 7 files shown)


ScottyPoi:
Made some changes to portalnetwork methods that need downstream updates.
Will address tomorrow

PR #383, opened 2022-10-29
Ignore List

ScottyPoi:
Adds to `PortalNetworkRoutingTable` a list of NodeId's to ignore.

Peers will be added to ignore list for 2 minutes upon eviction.

During this 2 minute period, any enr received in a NODES response will be ignored.

FINDNODES will also now listen for PONG response from unknown peers, and put unresponsive nodes in the same 2 minute ignore buffer.

This especially prevents against the rapid buildup of failing PING attempts when receiving the same unresponsive enr in multiple NODES responses.  

TODO:
Implement same logic in FINDCONTENT when CONTENT response is list of enrs

### DIFF
--- a/packages/portalnetwork/src/client/routingTable.ts
+++ b/packages/portalnetwork/src/client/routingTable.ts
@@ -5,11 +5,13 @@ export class PortalNetworkRoutingTable extends KademliaRoutingTable {
   private radiusMap: Map<NodeId, bigint>
   private gossipMap: Map<NodeId, Set<string>>
   private strikes: Map<NodeId, number>
+  private ignored: Set<NodeId>
   constructor(nodeId: NodeId) {
     super(nodeId)
     this.radiusMap = new Map()
     this.gossipMap = new Map()
     this.strikes = new Map()
+    this.ignored = new Set()
   }
 
   public setLogger(logger: Debugger) {
--- a/packages/portalnetwork/src/client/routingTable.ts
+++ b/packages/portalnetwork/src/client/routingTable.ts
@@ -19,13 +19,15 @@ export class PortalNetworkRoutingTable extends KademliaRoutingTable {
   }
 
   public strike = (nodeId: NodeId) => {
-    this.logger?.extend('STRIKE')(nodeId)
-    const strikes = this.strikes.get(nodeId) ?? 0
-    if (strikes > 1) {
+    let strikes = this.strikes.get(nodeId) ?? 0
+    strikes++
+    this.logger?.extend('STRIKE').extend(strikes.toString())(nodeId)
+    if (strikes > 2) {
       this.evictNode(nodeId)
-      this.strikes.delete(nodeId)
+      return
     }
-    this.strikes.set(nodeId, strikes + 1)
+    this.strikes.set(nodeId, strikes)
+    return strikes
   }
 
   public clearStrikes = (nodeId: NodeId) => {
--- a/packages/portalnetwork/src/client/routingTable.ts
+++ b/packages/portalnetwork/src/client/routingTable.ts
@@ -105,4 +105,13 @@ export class PortalNetworkRoutingTable extends KademliaRoutingTable {
       this.ignored.delete(nodeId)
     }, 120000)
   }
+
+  // Method for Protocol to check if Peer should be ignored.
+  // Mainly prevents self from continuing to PING dead enrs that we receive
+
+  public isIgnored = (nodeId: string) => {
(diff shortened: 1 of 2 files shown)


packages/portalnetwork/src/client/routingTable.ts
```diff
@@ -75,16 +79,39 @@ export class PortalNetworkRoutingTable extends KademliaRoutingTable {
...
     this.radiusMap.delete(nodeId)
     this.gossipMap.delete(nodeId)
-    if (enr) {
-      this.remove(enr)
+    this.strikes.delete(nodeId)
+  }
+
+  // Add node to ignored list for 2 minutes and then delete from ignored list
+
+  private ignoreNode = (nodeId: NodeId) => {
```
acolytec3, line 102: What do you think about having a single interval that checks the ignore list periodically and clears out ones that have passed a 2 minute mark rather than setting up a bunch of individual timeouts?  If we amass a large routing table, this could be a lot of open listeners.  We could even bake it into the bucket refresh listener, kind of a global table refresher method.  You could have the ignore list be a map or a sorted list of nodeIds -> timestamp and just prune the list for all timestamps greater than `Date.now() - 120000`.  That way, we don't have to worry about lots of dangling listeners in tests or elsewhere.

ScottyPoi: > You could have the ignore list be a map or a sorted list of nodeIds -> timestamp and just prune the list for all timestamps greater than Date.now() - 120000. That way, we don't have to worry about lots of dangling listeners in tests or elsewhere.

Clever. I'll do that.

packages/portalnetwork/src/subprotocols/contentLookup.ts
```diff
@@ -101,55 +92,48 @@ export class ContentLookup {
...
+                this.contacted.includes(decodedEnr.nodeId) ||
+                this.protocol.routingTable.isIgnored(decodedEnr.nodeId)
+              ) {
+                continue
               }
+              // Send a PING request to check liveness of any unknown nodes
               if (!this.protocol.routingTable.getValue(decodedEnr.nodeId)) {
-                this.protocol.routingTable.insertOrUpdate(decodedEnr, EntryStatus.Connected)
+                const ping = await this.protocol.sendPing(decodedEnr)
+                if (!ping) {
```
acolytec3, line 122: If we get a peerId during a lookup and that peer doesn't respond on the initial ping, do you think we should still keep them in the routing table or go ahead and evict them?  Seems like if we don't get an initial response from a node on the first ping, we should go ahead and boot them, right?

packages/portalnetwork/src/client/routingTable.ts
acolytec3: I think a slightly better way to do this would be to just create the ignore list as an array of tuples where we just push new tuples of `[timestamp, nodeId]` onto the end.  Then, we just do `splitIndex = ignoreList.findIndex((entry) => entry[0] > Date.now()-120000)` and then do `ignoreList = `ignoreList.slice(splitIndex)`.  What you're doing now is iterating over the entire ignoreList every time we do the bucket refresh but we should only need to find the first element with a timestamp < than 2 minutes ago and remove all items on the list prior to that.  Does that make sense?

acolytec3 (changes requested):
Left of a couple of comments on things we might want to consider changing.

packages/portalnetwork/src/subprotocols/contentLookup.ts
ScottyPoi: Yes, agreed.  That was the intention -- good catch :+1:

packages/portalnetwork/src/client/routingTable.ts
ScottyPoi: :+1: 
That works for me.  Made it so.

packages/portalnetwork/src/subprotocols/contentLookup.ts
ScottyPoi: Kind of wondering if we should bother with the 3 strikes thing at all, or just be savage about evicting nodes that don't PONG

acolytec3:
Probably doesn't hurt to just boot them. I know it's been said to bev more gracious by veggies implementers but I don't see why it matters. We get so many enrs from normal node activity we'll really refill any buckets that get emptied due to unresponsive nodes.

Nov 5, 2022 5:46:41 PM Scotty ***@***.***>:

PR #384, opened 2022-11-03
Database pruning and sublevels

ScottyPoi:
##### Introduces a new protocol for storing content in the database.
*portal-network-specs* defines a contentKey and contentId for each data type:
```
contentKey = selector + blockhash
contentId = sha256(contentKey)
```
Ultralight previously stored content in the database using `contentKey` hex strings as database keys 
 `contentId` was calculated to evaluate distance to a self or peer via `distance = nodeId XOR contentKey`

The reason XOR works for this purpose is the commutative property:  `A ⊕ B === B ⊕ A`
1. Every bytes32 `nodeId` has a unique XOR value with every possible bytes32 `contentId`.
2. Every `contentId` has a unique XOR value with every possible bytes32 `nodeId`

This PR creates a new key: `databaseKey` based on the unique distance from a client's `nodeId` to each piece of content.

`databaseKey = contentId ⊕ nodeId`

Storing content by `databaseKey` allows for iterating through database content according to distance, which will enable pruning by radius.

##### Introduces a new protocol for pruning content beyond a radius

A portal client's responsibility lies within it's radius.  When a client's content database approaches its limitations, it should first prune content beyond its radius, and then incrementally prune with a smaller radius to free up space.

LevelDB has a `.keys()` method that returns an iterator of keys in the database.
To parse contentKeys by distance would require iterating through the entire database and calculating distance.

Regardless of the order that keys were added, the keys are iterated in sorted order:  0x01, 0x02, 0x03, 0x04, 0x05...
`.keys()`  can also take a options regarding range.  So `db.keys({ gte: 0x03 })` would return:  0x03, 0x04, 0x05... 

If we do use `databaseKeys` to store content in the database, the content will sort itself by distance. The `.keys()` iterator will return in order of closest to farthest, and  `db.keys({ gte: radius })` will return an iterator of keys beyond the radius, which can be batch deleted in one operation.

##### Divides the Database into sections by sub-protocol

This pruning method is agnostic of the database values.  `contentId` is a hash of `contentKey`, so the content type cannot be derived from `databaseKey`.

This limitation becomes an issue as multiple `subprotocols` use the same database to store types of content with a large variation of content sizes.

`LevelDB` supports `sublevels`.   a `sublevel` instance behaves like a `LevelDB` instance.  Keys added to a sublevel are stored in the parent levelDB with an added prefix.  `db.keys()` will return all keys from all sublevels with their sublevel prefix.  `sublevel.keys()` will return all of the sublevel's keys, without the prefix.  

By opening a sublevel for each `protocol`, we can still allow users to provide a db if desired, while also enabling each subprotocol to prune its own section of the database independently of the rest.

### DIFF
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -146,7 +146,13 @@ export class PortalNetwork extends (EventEmitter as { new (): PortalNetworkEvent
     this.unverifiedSessionCache = new LRU({ max: 2500 })
     this.uTP = new PortalNetworkUTP(this.logger)
     this.refreshListeners = new Map()
-    this.db = new DBManager(this.logger, opts.dbSize, opts.supportedProtocols, opts.db) as DBManager
+    this.db = new DBManager(
+      this.discv5.enr.nodeId,
+      this.logger,
+      opts.dbSize,
+      opts.supportedProtocols,
+      opts.db
+    ) as DBManager
     opts.supportedProtocols = opts.supportedProtocols ?? []
     for (const protocol of opts.supportedProtocols) {
       switch (protocol) {
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -146,7 +146,7 @@ export class PortalNetwork extends (EventEmitter as { new (): PortalNetworkEvent
     this.unverifiedSessionCache = new LRU({ max: 2500 })
     this.uTP = new PortalNetworkUTP(this.logger)
     this.refreshListeners = new Map()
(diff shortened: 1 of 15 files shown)


acolytec3:
Is this one read for review?  It looks good on an initial skim but was going to play around with it a bit.

ScottyPoi:
> Is this one read for review? It looks good on an initial skim but was going to play around with it a bit.

It doesn't actually implement any checks on storage or protocol on what gets pruned when, but as far as methods for pruning, and a new system for database keys, yes it's ready for review.  

Measuring the size of the database is different in the browser vs. CLI - and will take some thought.  probably for a future PR

acolytec3 (approved):

PR #385, opened 2022-11-09
uTP -- Congestion Control refactor

ScottyPoi:
##  Refactor uTP congestion control 
1. `window`
2. `bitmask`
3. `selective ack`
4. `timestampDifferenceMicroseconds` / `reply_micro`
6. `rtt and rtt_var`
7. `timeout`
8. `delay => max_window`

#### Selective Ack implementation
  - Selective ack used to acknowledge a packet that has arrived out of order.
  - SelectiveAck `header.ackNr` corrected to be `socket.ackNr`
  - `socket.ackNr` only updated when *in-order* sequence is updated
    - out of order packets do not change socket.ackNr
  - SelectiveAck bitmask corrected to meet spec

#### `Bitmask` implementation
   - The 32 bit `Bitmask` in a `SelectiveAck header extension` represent 32 seqNr's beyond the break in the consecutive sequence.
     - `true` bits represent a `seqNr` that has been received
     - `false` bits represent a `seqNr` that has not been received
   -  `Write` sockets parse this header extension to understand which SeqNr's have been missed
   
 
####  Window Size
  - max_window is dynamically increased or decreased to control send rate
  - cur_window is increased when data packets are sent
  - cur_window is decreased when data packets are acked
  - A Socket not send another packet if `cur_window + packet_size > max_window`

#### Timestamps
  - Bytes32Timestamp() altered for accuracy
  - A microsecond timestamp is recorded when packets are received
    - `socket.reply_micro = Bytes32Timestamp()`
  - A microsecond timestamp is recorded when a packet is created
    - `packet.header.timestampMicroseconds = Bytes32Timestamp()`
    - `packet.header.timestampDifferenceMicroseconds = socket.reply_micro`

#### rtt / rtt_var
  - Sockets keep track of RTT (round trip time) between sending a DATA packet and receiving the ACK
  - Sockets also track the variation in RTT times.
  - Sockets update these values by a particular formula.
  - Sockets use this formula to then calculate `TIMEOUT` 

#### base_delay
  - Sockets compare a packet's timestamp against the time received to calculate `delay`
  - Sockets compare this delay value to a `base_delay`
  - Sockets then increase or decrease max_window based on delay times.

#### Timeout / Throttle
  - Sockets set a dynamic time limit to wait for responses based on recent RTT and RTT_VAR
  - When a socket times out - it throttles its send rate by controlling max_window

### Added to utp:
  1. socket.generateSelectiveAckBitMask(): Uint8Array
    - Generates a bitmask from socket.ackNr and socket.ackNrs
  2. PortalNetworkUtp.bitmaskToAckNrs(bitmask: Uint8Array, ackNr: number): number[]
    - Takes the bitmask from a selective ack packet header extension and returns list of ackNrs
  3. PortalNetworkUtp._handleSelectiveAck()
  4. socket.updateRTT
  5. socket.updateDelay
  6. socket.throttle
  7. socket.timeoutCounter
  8. utp.closeRequest() & socket.close()

#### Update protocol to close uTP socket


ScottyPoi:
**ack === "Acknowledge that we received the packet"**
**Ack === ST_STATE packet used to *ack***
**Selective Ack === *Ack* packet with `header.extension ===  1`, and  `header-extension`** 
  - `SelectiveAck.header.ackNr` === *highest index of **consecutive** ST_DATA.header.seqNr received*
  - `Header-Extension`
    * `type`
    * `length`
    * `bitmask`
      * *bitmask* is a 4byte data structure containing a list of the sequence numbers received beyond the last in-order sequence number.   
  - i.e. 
    - *socket receives seqNr ...4...5...6...7... in order, but packets 9 and 10 before 8*
    - *ack* arrival of packet 9 with `SelectiveAck`
      - ackNr: 7
      - bitmap: [9]
    - *ack* arrival packet 10 with `SelectiveAck`
      - ackNr: 7
      - bitmap: [9, 10]
    - *ack* arrival of packet 8 with `Ack`
      - acknNr: 10

- When to send *Selective Ack* vs *Ack*?
  - Ack 
    - To acknowledge a ST_SYN packet 
    - To acknowledge the *highest index* of `ST_DATA.header.seqNr` received in an *unbroken* consecutive sequence
        - Packet arrives with next `header.seqNr` in unbroken consecutive order
          - 5...6...7...**8**
          - Ack: `ST_STATE.header.ackNr = 8`
        - Packets arrives that makes sequence of received `ST_DATA.header.seqNr` consecutive again
          - 5...6...7..._...9...10...**8**
          - Ack: `ST_STATE.header.ackNr = 10`
    - To acknowledge a FIN packet once all ST_DATA packets have been accounted for.
        - Wait to *ack* until all `seqNr` up to `ST_FIN.header.seqNr` have been *acked*
          - Ack: `ST_STATE.header.ackNr = ST_FIN.header.seqNr`
  - Selective Ack
      - To acknowledge a ST_DATA packet when consecutive sequence of `header.seqNr` has been broken
        - Packet arrives with `header.seqNr` ahead of  the highest index `seqNr` in the *unbroken consecutive chain*

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
```diff
@@ -178,23 +178,20 @@ export class UtpSocket extends EventEmitter {
...
     this.seqNr = this.seqNr + 1
-    this.ackNr = packet.header.seqNr
     if (!this.reader) {
       this.reader = new ContentReader(this, packet.header.seqNr)
     }
+    this.ackNrs.push(packet.header.seqNr)
     await this.reader.addPacket(packet)
     if (expected) {
-      this.ackNrs.push(this.ackNr)
+      this.ackNr = Math.max(...this.ackNrs)
```
ScottyPoi, line 183: This should actually be set to the highest in-order ackNr, not necessarily the highest received.

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/index.ts
```diff
@@ -402,24 +400,17 @@ export class PortalNetworkUTP extends BasicUtp {
         throw new Error('Missing content keys')
       }
       keys.forEach((k, idx) => {
```
ScottyPoi, line 403: change to `for (const [idx, k] of keys.entries()) {...}`

ScottyPoi:
TODO:
1. Sending socket should resend a data packet if selective ack is >3 packets ahead of socket.ackNr
2. Sending socket should slow data stream if rtt is slow.

ScottyPoi:
#### rtt, rtt_var, and timeout

* Every packet that is ACKed...should be used to update an rtt (round trip time) and rtt_var (rtt variance) measurement. 
* The rtt and rtt_var is only updated for packets that were sent only once
* rtt and rtt_var are calculated by the following formula, every time a packet is ACKed:
  * ```
delta = rtt - packet_rtt
rtt_var += (abs(delta) - rtt_var) / 4;
rtt += (packet_rtt - rtt) / 8;
```
* The default timeout for packets associated with the socket is also updated every time rtt and rtt_var is updated. It is set to:
  * ```
timeout = max(rtt + rtt_var * 4, 500);
```
  * Where timeout is specified in milliseconds. i.e. the minimum timeout for a packet is 1/2 second.

packages/portalnetwork/src/subprotocols/history/contentManager.ts
```diff
@@ -132,11 +134,28 @@ export class ContentManager {
...
+              { hardforkByBlockNumber: true }
+            ).number
+          }`
+        )
         this.history.client.db.put(
           getHistoryNetworkContentKey(HistoryNetworkContentTypes.Receipt, fromHexString(hashKey)),
           toHexString(value)
         )
-        break
+        return
```
acolytec3, line 158: Do we not gossip receipts?

acolytec3 (changes requested):

packages/portalnetwork/src/wire/utp/Packets/Packet.ts
```diff
@@ -25,29 +26,27 @@ export class Packet {
...
+            timestampDifferenceMicroseconds: BigNumber.from(buffer.subarray(8, 12)),
+            wndSize: BigNumber.from(buffer.readUInt32BE(12)),
             seqNr: buffer.readUInt16BE(16),
             ackNr: buffer.readUInt16BE(18),
           },
-          buffer.subarray(22, 22 + size)
+          bitmask
         ),
-        payload: buffer.subarray(22 + size),
+        payload: Uint8Array.from([]),
```
acolytec3, line 49: Just to confirm my understanding, there's no payload with a selective Ack?

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/index.ts
```diff
@@ -33,17 +40,27 @@ export interface INewRequest {
...
+    this.openContentRequest = new Map()
     this.working = false
   }
 
+  closeRequest(nodeId: string) {
+    this.logger.extend('CLOSING')(`Closing uTP request with ${nodeId} due to failed connection`)
+    const key = Object.keys(this.openContentRequest).find(
+      (k) => k.slice(0, 5) === nodeId.slice(0, 5)
+    )
+    !key && this.logger.extend('CLOSING')('Cannot find request')
```
acolytec3, line 59: This is a small thing but can we make this something like below?  Otherwise, we're doing multiple comparisons when really only one is needed (whether or not `key` is defined or not)
```js
if (key === undefined) {
    this.logger.extend...
}
else {
    this.openContentRequest.get...
    this.openContentRequest...
}
```

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/index.ts
```diff
@@ -339,27 +339,70 @@ export class PortalNetworkUTP extends BasicUtp {
...
+    const acked = ackNrs.find((a) => !request.socket.ackNrs.includes(a))
+    request.socket.logger(
+      `ST_STATE (SELECTIVE_ACK) received with ackNr: ${
+        packet.header.ackNr
+      }, and a bitmask referencing ackNrs: ${ackNrs}.  Packet acks DATA packet seqNr: ${acked}.  Receive socket still waits for seqNr: ${
+        packet.header.ackNr + 1
+      }`
+    )
+    if (acked) {
+      // request.socket.rtt = request.socket.reply_micro - request.socket.outBuffer.get(acked)!
```
acolytec3, line 387: Do we need this?

packages/portalnetwork/src/wire/utp/Protocol/write/ContentWriter.ts
```diff
@@ -29,31 +28,36 @@ export default class ContentWriter {
...
+      if (this.socket.cur_window + DEFAULT_PACKET_SIZE <= this.socket.max_window) {
+        bytes = this.dataChunks[this.seqNr] ?? []
+        this.sentChunks.push(this.seqNr)
+        this.socket.logger(
+          `Sending ST-DATA ${this.seqNr - this.startingSeqNr + 1}/${chunks} -- SeqNr: ${
+            this.socket.seqNr
+          }`
+        )
+        await sendDataPacket(this.socket, bytes)
+        this.seqNr = Math.max(...this.sentChunks) + 1
```
acolytec3, line 43: Is it possible these chunks get pushed into the `sentChunks` out of order?  If we're pushing the `seqNr` sequentially into the array, shouldn't we be able to just look at the last one?

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
```diff
@@ -145,52 +151,69 @@ export class UtpSocket extends EventEmitter {
...
-    this.ackNr = packet.header.seqNr
     if (!this.reader) {
       this.reader = new ContentReader(this, packet.header.seqNr)
     }
+    this.ackNrs.push(packet.header.seqNr)
     await this.reader.addPacket(packet)
     if (expected) {
-      this.ackNrs.push(this.ackNr)
+      this.ackNr =
+        this.ackNrs.sort((a, b) => a - b).find((n, i) => this.ackNrs[i + 1] !== n + 1) ??
```
acolytec3, line 206: This feels like a lot of work to find the right ack number if we're having to sort the array every single time.  Is there any difference in result between the two parts of this expression?

ScottyPoi:
Thanks for looking it over.
This ended up being so much more involved than expected, so i'm sure i made mistakes along the way

packages/portalnetwork/src/wire/utp/Packets/Packet.ts
ScottyPoi: correct, there is a header-extension with a bitmask, but no payload

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/index.ts
ScottyPoi: I was trying to find a way for PortalNetwork to signal to uTP to close a request when the discv5 connection to the peer has been interrupted, using just the PeerId.  
```
      try {
        await this.sendPortalNetworkMessage(enr ?? peerId, msg, protocolId, true)
      } catch {
        this.uTP.closeRequest(peerId)
      }
```
Looking at it again, though, I could just deserialize `msg` and get the connectionIds, and the added search.

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
ScottyPoi: OK help me think through this...

* `this.ackNr` only changes when the in-order sequence grows.
* `this.ackNrs` does update even with out of order packets, so `this.ackNrs` can end up unsorted.
* An "expected" packet means the next in-order packet is being acked, whether it is at the tip of the sequence or in the middle.  Then this.ackNr changes to the last of the in-order ackNrs, which could still be in the middle of the array of ackNrs.

packets are coming in like:
5...6...7...9...11...
`this.ackNrs = [5, 6, 7, 9, 11]`

9 and 11 were out of order, so got selective-acks, this.ackNr doesn't change.
`this.ackNr = 7`

then we get packet 8, and respond with a regular ACK, with an ackNr that references the new tip of the in-order sequence.
so `this.ackNrs = [5, 6, 7, 9, 11, 8]`
and `this.ackNr` needs to change to *9*, which is the highest of the in-order seqNrs, but it's not the largest number, or the last in the array.
```
        this.ackNrs.sort((a, b) => a - b).find((n, i) => this.ackNrs[i + 1] !== n + 1) ??
```
 
any other ideas on how to find the new ackNr?

packages/portalnetwork/src/wire/utp/Protocol/write/ContentWriter.ts
ScottyPoi: It could happen if we have to resend a lost packet, and then skip back ahead to where we were.

Preventing the same seqNr from being stored twice is easy though, so unless I'm forgetting something then yes, we should just be able to look at the last seqNr in the array

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/index.ts
ScottyPoi: nope

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
acolytec3: I'm not sure if this is better or not but what if we change `this.ackNrs` to a `Set`?  Then, the above would change to something like:
```js
let i = this.ackNr + 1
let foundNextAck = false
while (!foundNextAck) {
  if (this.ackNrs.has(i)) {
    i++
  } else {
     foundNextAck = true
     this.ackNr = i
  }
}
```
Then, we're only iterating over the ack values greater than `this.ackNr` and not having to do a bunch of sorting first.  I'm not 100% sure this is any more efficient but it makes sense in my head that it would be faster.

ScottyPoi: Shouldn't need to.
```
const list = [1, 2, 3]
// list.length === 3
list[4] = 5
// list === [1,  2, 3, undefined, 5]
// list.length === 5
// list[3] === undefined
```

ScottyPoi: no pushing involved.

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/index.ts
ScottyPoi: actually...yes.

acolytec3 (approved):
LGTM

PR #390, opened 2022-11-21
Fix handleFindContent

ScottyPoi:
Network content lookups were broken.
Peers would respond with FOUNDCONTENT for content keys not in their DB.

The reason was that during `protocol.handleFindContent`, when we check our local DB for content, we would only check if the return value was `undefined`.  Sometimes this return value is an empty Uint8Array, so we would return this empty array is if it was the requested content.

Fixed by adding a check for `value.length === 0` after `protocol.findContentLocally`

### DIFF
--- a/packages/portalnetwork/src/subprotocols/protocol.ts
+++ b/packages/portalnetwork/src/subprotocols/protocol.ts
@@ -449,7 +449,7 @@ export abstract class BaseProtocol {
 
     const lookupKey = serializedContentKeyToContentId(decodedContentMessage.contentKey)
     const value = await this.findContentLocally(decodedContentMessage.contentKey)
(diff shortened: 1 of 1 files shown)


ScottyPoi:
@holgerd77 
I think I solved an issue with ONE line of code! :1st_place_medal:

holgerd77:
> @holgerd77 I think I solved an issue with ONE line of code! 🥇

Haha, achievement unlocked! 💯 😜 🙂

acolytec3:
I'm fine merging this though do have a question.  Why are we ever returning empty arrays from the DB?  Do we have a bug and sometimes put empty content in the DB?

acolytec3 (approved):
LGTM.

PR #395, opened 2022-12-07
Updates to CLI ReadMe

ScottyPoi:
* Updates list of JSON-RPC methods
* Adds quickstart instructions for Devnet
  * Adds npm script `npm run devnet`
  * Adds --port variable to set the "starting port" number
    * i.e. `npm run devnet -- --numNodes=3 --port=8545`
    * starts `3` nodes on `8545`, `8546`, and `8547`
* Adds Basic example using **devnet** of 2 nodes
  * via command-line JSON-RPC
    * `discv5_nodeInfo` => `portal_historyAddBootNode` => `ultralight_addBlockToHistory` => `eth_getBlockByHash`
  * via typescript test scripts
    * creates `scripts/sampleTest.ts` to perform same sequence
*  Fix bug in eth_getBlockByHash JSON-RPC call
   *  Would return `Block not found` message while successfully finding block. (aka **LYING**)


ScottyPoi:
Merging to avoid clashes with other open PR's.

No big changes, just updates

Issue #398, opened 2023-02-03
uTP `contentWriter` race condition


acolytec3:
The [`eth_getBlockByNumber`](https://github.com/ethereumjs/ultralight/blob/06d69100116528758f0df200770261375fdbf24b/packages/portalnetwork/test/integration/integration.spec.ts#L207) integration test fails occasionally with this sort of issue:
![image](https://user-images.githubusercontent.com/17355484/216639322-be547534-f75d-44c6-be38-4b548ad53779.png)

The `contentWriter.start` method seems to occasionally get stuck if the `cur_window` and `max_window` values aren't updated on the correct tick.  We could explore reworking `contentWriter.start` away from recursively calling itself to use something like:

```
for (const chunk of chunks) {
   await sendPacket
```
or
```
for await (const chunk of chunks) {
   sendPacket
   next()
```

ScottyPoi:
I don't think it can be that simple and follow the congestion control rules of uTP.

we can't just stream out packets.  There is supposed to be a limit of 3 in-flight packets, meaning data packets that have been sent but the ack not yet received. 

we're supposed to recalculate the round-trip time stuff for each packet, and use that metric to set a timeout function that is supposed to consider a packet lost if it is taking longer than expected.

we have to watch for ack / selective ack, and use that to stay no more than 3 packets ahead, while also resending packets that may have been lost.

I agree that the writer could be written differently, or better, or optimized, but it's never going to be as simple as `for (chunk of chunks) { sendChunk }`

PR #400, opened 2023-02-03
uTP improvements

ScottyPoi:
DRAFT/WIP

PR to address issue #398 and other uTP related improvements or fixes.


ScottyPoi:
This grew into a larger refactor / reorg of the uTP implementation, including:

* Improved Typing for Packets
  * Packet<T>, Header<T>, IPacketOpts<T>, IHeaderOpts<T>, etc.
* Separates CC protocol into `CongestionControl` class.
* Separates Packet management into `PacketManager` class.
* Reorg and consolidation of uTP directory into `Utils/`, `Packets/`, `Socket/`, and `Protocol/`
* Removes a lot of circularity 
  * e.g. removing the need for a socket pass itself as a parameter in `socket.reader = new ContentReader(socket)`
* Improved write / send protocol, addressing #398

ScottyPoi:
Structurally, things are more logical, i think.  This is a birdseye sketch:
```
* PortalNetwork
  * .handle(discv5_packet)
  * .send(discv5_packet)
  * uTP
    * .handleNewRequest()
    * .openUtpSocket()
    * Map<requestKey, contentRequest>
    * ContentRequest[ ]
      * ContentRequest
        * contentKey(s), requestCode
        * Socket
          * content
          * .send(Packet<T>)
          * .handle(Packet<T>)
          * Reader/Writer
            * .chunk(content) / .compile(chunks)
            * Map<seqNr, chunk>
          * Packet Manager
            * .create() => Packet
              * .create() => Header
            * CongestionControl
              * .updateRTT()
              * .throttle()
              * .updateWindow()
              * .updateDelay()
```

ScottyPoi:
I didn't find a way to use an AsyncGenerator* write function, but I did remove the recursion from the writer.start() method.  

write() and start() are now separate methods, and write() does not call itself recursively anymore.  the writer simply chunks the data, and stores a map of seqNr's to data chunks.  write() is called by the socket as state packets are handled.

"Send" methods now emit "Send" events, and return a Promise that resolves when a "Sent" event is triggered by PortalNetwork receiving the corresponding talkResp.

ScottyPoi:
I did find a bug in the congestion control that may have contributed to the issues in #398.

In `.updateDelay`, i was using a dynamic delay target, instead of the constant of 100ms from the spec.  This contributed to the tendency for the delay and window values to go to 0 or INFINITY.  This should not happen anymore.

ScottyPoi:
I think there may still be ways to simplify / optimize the sending/handling of packets.  

In striving towards smaller files, Socket, and PortalNetworkUTP, while simplified, are both still ~400 lines each.

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
```diff
@@ -87,213 +121,198 @@ export class UtpSocket extends EventEmitter {
...
-  async sendFinPacket(packet: Packet): Promise<Buffer> {
+  async sendFinPacket(): Promise<void> {
+    const packet = this.createPacket<PacketType.ST_FIN>({ pType: PacketType.ST_FIN })
     this.finNr = packet.header.seqNr
-    return await this.sendPacket(packet, PacketType.ST_FIN)
+    await this.sendPacket<PacketType.ST_FIN>(packet, PacketType.ST_FIN)
+  }
+
+  async sendDataPacket(bytes: Uint8Array): Promise<Packet<PacketType.ST_DATA>> {
+    await this.packetManager.congestionControl.canSend()
```
acolytec3, line 177: Don't we need to check the result of this call before sending a packet?  Out of curiosity, how does the usage of this function differ from the `canSend` event?

packages/portalnetwork/src/wire/utp/Socket/congestionControl.ts
```diff
@@ -0,0 +1,105 @@
...
+    const scaledGain = MAX_CWND_INCREASE_PACKETS_PER_RTT * delayFactor * windowFactor
+    const new_max = this.max_window + scaledGain > 0 ? this.max_window + scaledGain : 0
+    this.max_window = new_max
+  }
+  updateWindow() {
+    const inFlight = this.outBuffer.size
+    this.cur_window = inFlight * DEFAULT_PACKET_SIZE
+    this.logger(`cur_window: ${this.cur_window} bytes in flight`)
+    if (this.cur_window + DEFAULT_PACKET_SIZE <= this.max_window) {
+      this.emit('canSend')
```
acolytec3, line 102: What is this event doing in the current version of the code?  I don't see anywhere in the code where this event is listened for other than the `canSend` method and it's commented out there.

acolytec3 (changes requested):
I haven't reviewed every line in depth but this definitely is starting to make more sense.  Left some specific feedback around `canSend` the method versus `canSend` the event that should be addressed.  Also, can we generally clean up all the commented out code and remove unused deps?  Once the one specific question is cleared up and we address the other stuff, this should be merge ready methinks.

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
```diff
@@ -1,84 +1,118 @@
...
-import ContentWriter from '../Protocol/write/ContentWriter.js'
-import ContentReader from '../Protocol/read/ContentReader.js'
-import { BasicUtp } from '../Protocol/BasicUtp.js'
-import { sendAckPacket, sendSynAckPacket } from '../Packets/PacketSenders.js'
+import ContentWriter from './ContentWriter.js'
+import ContentReader from './ContentReader.js'
+import { PacketManager } from '../Packets/PacketManager.js'
 import { BitArray, BitVectorType } from '@chainsafe/ssz'
 
+interface ReadSocketOptions extends UtpSocketOptions {
```
acolytec3, line 19: Do we need these two interfaces?  If not, can we remove them?

packages/portalnetwork/src/wire/utp/Socket/congestionControl.ts
ScottyPoi: ah, yes.  I wasn't sure how to write a test for it, so i had left it commented out with a shortcut.  
real solution in place now.  will explain in PR comment

ScottyPoi:
### canSend()

- uTP:
  - TLDR: 
    - We track the round trip time of data packets, and use it as a metric to throttle the send rate.
    - Sometimes this means waiting to send the next packet 


- `socket.sendDataPacket()`
  - calls `await this.congestionControl.canSend()` 
- `.canSend()`
  - checks the send rate/window
  - if the socket can send
    - returns `true` 
  - if the socket can't send
    - returns a `Promise` which resolves when a `canSend` event is emitted by `.updateWindow()`
-  socket.congestionControl.updateWindow() is called when packets are received
  - along with `.updateDelay()` and `.updateRTT()`
  - emits a `canSend` event if `this.max_window >= this.cur_window + DEFAULT_PACKET_SIZE`

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/index.ts
```diff
@@ -83,139 +70,53 @@ export class PortalNetworkUTP extends BasicUtp {
...
+      protocolId: ProtocolId.HistoryNetwork,
+      requestCode,
+      socket: this.createPortalNetworkUTPSocket(requestCode, peerId, sndId, rcvId, content),
+      socketKey: createSocketKey(peerId, sndId, rcvId),
+      content: params.contents ? params.contents[0] : undefined,
+      contentKeys,
+    })
+    this.openContentRequest.set(newRequest.socketKey, newRequest)
+    this.logger(`Opening request with key: ${newRequest.socketKey}`)
+    await newRequest.init()
```
acolytec3, line 119: Do we need to do anything with the requestCode returned by this method?

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
```diff
@@ -87,213 +116,196 @@ export class UtpSocket extends EventEmitter {
...
-  async sendFinPacket(packet: Packet): Promise<Buffer> {
+  async sendFinPacket(): Promise<void> {
+    const packet = this.createPacket<PacketType.ST_FIN>({ pType: PacketType.ST_FIN })
     this.finNr = packet.header.seqNr
-    return await this.sendPacket(packet, PacketType.ST_FIN)
+    await this.sendPacket<PacketType.ST_FIN>(packet, PacketType.ST_FIN)
+  }
+  async sendDataPacket(bytes: Uint8Array): Promise<void> {
+    this.state = ConnectionState.Connected
+    await this.packetManager.congestionControl.canSend()
```
acolytec3, line 166: Will this ever return false?  If not, can we skip the boolean and just await it?

packages/portalnetwork/src/wire/utp/Socket/congestionControl.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.rtt = 1000
+    this.rtt_var = 0
+    this.timeout = 1000
+    this.baseDelay = { delay: 0, timestamp: 0 }
+    this.ourDelay = 0
+    this.sendRate = 0
+    this.outBuffer = new Map()
+  }
+
+  async canSend(): Promise<boolean> {
```
acolytec3, line 39: Just thinking this through.  Should we have some sort of timeout on this so if the `canSend` event isn't emitted within say 3 seconds, we abort the promise and close the socket?  Seems like a memory leak waiting to happen if we leave these promises just sitting waiting if the remote peer doesn't ACK some of the in flight packets because it hangs up/goes offline/etc.

acolytec3 (approved):
Tests fixed, logic simplified.  LGTM!

PR #401, opened 2023-02-06
Add new header with proof type

acolytec3:
Adds the new header with proof types from [this pr](https://github.com/ethereum/portal-network-specs/pull/177).

TODO:
- [ ] Add a hardcoded full master accumulator so we can generate and verify premerge proofs
- [ ] Add tests for the premerge header proofs

----

#### HeaderWithProof

- Updates History Network content type `BlockHeader` to the `HeaderWithProof` content type (still called `BlockHeader`)

#### BaseProtocol / HistoryProtocol

Changes the constructor for `BaseProtocol` to consume a set of bound methods instead of passing the PortalNetwork class itself as a parameter.

Updates `BaseProtocol` and `HistoryProtocol` to call methods directly.

from: 
```javascript
class PortalNetwork {
... 
  constructor() {
...
    this.protocols.set('0x500b', new HistoryProtocol(this))
  }
}
```
to
```javascript
class PortalNetwork {
... 
  constructor() {
...
   p = new HistoryProtocol({
            sendMessage: this.sendPortalNetworkMessage.bind(this),
            sendResponse: this.sendPortalNetworkResponse.bind(this),
            findEnr: this.discv5.findEnr.bind(this.discv5),
            put: this.db.put.bind(this.db),
            get: this.db.get.bind(this.db),
            handleNewRequest: this.uTP.handleNewRequest.bind(this.uTP),
            prune: this.db.prune.bind(this.db, protocol),
            ENR: this.discv5.enr,
            metrics: this.metrics,
          })
          this.protocols.set('0x500b', p)
  }
}
```

#### addContentToHistory / ContentManager

- Removes the `ContentManager` and `ReceiptManager` attributes from `HistoryProtocol`.
- Replaces `addContentToHistory()` with `history.store()`
- Special handling for `HeaderWithProof` and `BlockBody` still part of `HistoryProtocol`
  - validates the header/proof before adding to DB
  - validates the block body with corresponding header before adding to DB
  - seeks header from the network if needed to validate block body (won't add without header)
  - generates and gossips block `Receipt` content

#### DBManager / Prune()

- Moves pruning logic to global DBManager.
- `protocol.emit('prune', newRadius)`
- client: `for (protocol of protocols) { protocol.on('prune', (newRadius) => this.db.prune(protocol, newRadius))  }`

#### ReceiptManager

- Replaces ReceiptManager class with a set of exported methods for generating receipts.
- Retrieval and Storage of Blocks and Receipts handled in HistoryProtocol class
- 

#### HeaderAccumulator / ValidateHeaderProof

- Updates HeaderAccumulator class with "Final" version of master accumulator epoch hashes.
- Updates `Proof` creation and validation for new `HeaderWithProof` type
 

#### Type / Interface name  changes

- Detailed in comment
- Removes the `HistoryNetwork` tag from several type and interface names
  - Reduced the code size in a meaningful way
- e.g. `getHistoryNetworkContentKey` => `getContentKey`


ScottyPoi:
portal network integration test passes locally.  Though sometimes takes a restart.  Issue is in uTP congestion control timer.

Fix for that in PR #400

packages/portalnetwork/src/subprotocols/history/contentManager.ts
```diff
@@ -41,8 +38,9 @@ export class ContentManager {
 
     switch (contentType) {
       case HistoryNetworkContentTypes.BlockHeader: {
+        const headerWithProof = BlockHeaderWithProof.deserialize(value)
```
acolytec3, line 41: Should the `value` passed into `addContentToHistory` move from being an RLP serialized header or block body or receipt to being an SSZ serialized object?  Looks like a variety of tests around this are failing because this function is now expecting an SSZ serialized value.  I don't think there's a right or wrong answer but not sure which way makes more sense.  We originally built this method as a convenience method to handle blocks and headers imported via RPC but it gets murky once you add in epoch accumulators.  I'm not sure I have a clear view on this but if we want outside users to consume our code, I kinda think we should keep the input (at least at the protocol level) as an RLP encoded header or block.  It might make sense to rewrite the `contentManager` method to expect SSZ serialized portal network objects and leave `addContentToHistory` as a convenience method exposes by the provider or downstream consumer.

ScottyPoi: >  It might make sense to rewrite the contentManager method to expect SSZ serialized portal network objects and leave addContentToHistory as a convenience method exposes by the provider or downstream consumer.

I agree with this.  I don't think it makes sense to separate the header and the proof once validated.    

'addContentToHistory' can be a convenience method, and the actual DB storage and retrieval can be simpler. :thumbsup

ScottyPoi:
Added to this PR:
`history.contentManager`, and `history.addContentToHistory` have been removed.

replaced with `history.store()` and `history.retrieve()` , as well as `history.addBlockBody`, which will validate the block body using the corresponding header, and save the `block receipts`

Some names were shortened to drop the `HistoryNetwork` label.
`HistoryNetworkContentTypes` becomes `ContentTypes`
`HistoryNetworkContentKey` becomes `ContentKey`
`HistoryNetworkContentId` becomes `ContentId`
`getHistoryNetworkContentKey` becomes `getContentKey`
`getHistoryNetworkContentId` becomes `getContentId`

unused `HistoryNetworkContentType` types were deleted.

acolytec3:
Is this ready for review?

ScottyPoi:
Yeah -- 
Not sure why the integration test timeout happened.  Passes locally.

ScottyPoi:
I can do a better write-up for it, first, since there is more here than just the new content type
In tinkering with addContentToHistory, I made bigger changes that probably could have been done over a few PR's instead.

packages/portalnetwork/src/subprotocols/history/history.ts
```diff
@@ -1,113 +1,97 @@
...
-  public init = async () => {
-    this.client.uTP.on('Stream', async (selector, blockHash, content) => {
-      if (selector === HistoryNetworkContentTypes.EpochAccumulator) {
-        blockHash = toHexString(
-          EpochAccumulator.hashTreeRoot(EpochAccumulator.deserialize(content))
-        )
-      }
-      await this.addContentToHistory(selector, blockHash, content)
+
+  public init = async () => {}
```
acolytec3, line 78: I think you can safely get rid of this and then the abstract method in the base class.

packages/portalnetwork/src/subprotocols/protocol.ts
```diff
@@ -29,23 +28,64 @@ import {
...
   private checkIndex: number
-  protected abstract logger: Debugger
+  abstract logger: Debugger
   abstract protocolId: ProtocolId
   abstract protocolName: string
-  public client: PortalNetwork
-  constructor(client: PortalNetwork, nodeRadius?: bigint, metrics?: PortalNetworkMetrics) {
-    this.client = client
+  public ENR: ENR
+  handleNewRequest: (request: INewRequest) => Promise<ContentRequest>
```
acolytec3, line 63: I'm not really sure how I feel about this one.  While I sort of get why you pulled everything out like this, it still obscures the implementation details of all these methods and makes them hard to find.  I guess I could go either way with it at the end of the day if you have a compelling reason for abstracting all these methods out but it feels weird to me to have a bunch of methods with no implementation details available within the protocol.  The main case I can think of for doing it this way is if we're going to build separate client implementations that interact with these protocols and I don't think that'll ever be the case (unless we completely rip out the guts of the client and replace the whole thing with a libp2p client - for example).  That, to be honest and despite my preference, is probably a long way off.  If it were just me, I'd keep everything as it was before so we have access to the full client as needed the next time we're building out a subprotocol.  I dunno, will have to think more on it.

ScottyPoi: OK -- I can live with rolling back that part.  Not really in the scope of this PR anyway.

packages/portalnetwork/src/subprotocols/history/history.ts
ScottyPoi: check :heavy_check_mark:

PR #402, opened 2023-02-23
State Network Scaffolding

acolytec3:
# TODO
- [ ] Finish SSZ types for storage trie proof and contract bytecode
- [ ] Write tests to generate proofs, serialize, and then deserialize the types
- [ ] Write utils to create content keys/contentIds
- [ ] Implement stub functions `sendFindContent`, `findContentLocally`, `store`
- [ ] Write some tests to pipe in some content and then gossip it around a devnet

### DIFF
--- a/packages/portalnetwork/src/subprotocols/state/types.ts
+++ b/packages/portalnetwork/src/subprotocols/state/types.ts
@@ -1,18 +1,25 @@
-import { ByteVectorType, UintBigintType } from '@chainsafe/ssz'
-import { Witnesses } from '../history'
+import {
+  ByteListType,
+  ByteVectorType,
+  ContainerType,
+  ListCompositeType,
+  UintBigintType,
+} from '@chainsafe/ssz'
 
-export enum ContentType {
+export enum StateNetworkContentType {
   AccountTrieProof = 0,
   ContractStorageTrieProof = 1,
   ContractByteCode = 2,
 }
 
 /* ----------------- Types ----------- */
-export type Bytes32 = Uint8Array
+type Bytes32 = Uint8Array
 export type Bytes20 = Uint8Array
 export type Uint256 = BigInt
 export type Address = Bytes20
 export type StateRoot = Bytes32
+export type StateRootHex = string
+export type MPTWitnessNode = Uint8Array
 export type AccountTrieProofKey = {
   address: Address
   stateRoot: StateRoot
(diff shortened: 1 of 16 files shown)


packages/portalnetwork/src/subprotocols/state/types.ts
```diff
@@ -0,0 +1,96 @@
...
+export const ContractStorageTrieProofType = new ContainerType({
+  witnesses: MPTWitnessesType,
+  data: Bytes32Type,
+})
+
+export const ContractByteCodeKeyType = new ContainerType({
+  address: AddressType,
+  codeHash: Bytes32Type,
+})
+export const ContractByteCodeType = {
```
ScottyPoi, line 94: this is off.

this should either be:
```typescript
export const ContractByteCodeType = new ListBasicType(new UintBigintType(1), 24576)
```
or
```typescript
export const ContractByteCodeType = new ContainerType({ bytecode: ListBasicType(new UintBigintType(1), 24576) })
```

ScottyPoi: I don't see in the spec where / if this SSZ object is defined as a list or a container

ScottyPoi: and actually, we can use the `ByteList` alias here.

`... new ByteList(24576)`

ScottyPoi:
@acolytec3 let's think this through --

does it make sense to store all of the state content as ready for delivery like we do the history content?

in this case the majority of content includes proofs, and the proofs are much larger (RLP amiright? :roll_eyes: ).
also we may end up storing individual witness hashes over and over again, in lots of different lists.

would it instead make sense for the portal client to use the ethjs statemanager to maintain a history of checkpoints from tries with the content we care about / want to serve?

ScottyPoi:
i haven't fully thought it though...
it feels like maybe we could store a lot less content in the db that way, for the expense of recomputing proofs for content that actually gets requested.
but i'm not sure if that actually makes sense with the distribution of contentId's or not.

acolytec3:
> i haven't fully thought it though... it feels like maybe we could store a lot less content in the db that way, for the expense of recomputing proofs for content that actually gets requested. but i'm not sure if that actually makes sense with the distribution of contentId's or not.

That's a really interesting idea. Just an initial thought.  It in theory should be a fair bit smaller than maintaining all the proofs (which can easily get into the megabytes in size).  That said, you also run the risk of not being able to prune the db.  Once we have a state trie.  We can't easily delete nodes from it (unless we do it manually by snipping from the trie directly versus just calling `stateManager.deleteAccount`).  If you use the state manager to delete nodes from the trie, you end up changing the state root and we definitely don't want to do that.  What I do know is that we can just dump proof nodes into the trie.  We did this when back when I was trying to build trie-backed version of the ethers state manager where we would call `getProof` and then just fill in the the parts of the trie that we got from the proof nodes.  This might be a viable thing for us to do though it makes it hard to really enforce the "radius" bits of the protocol network since the underlying DB is storing things differently.  We could even hack around that though by implementing our own DB wrapper for level DB that uses your DB manager underneath so we could still prune the db without affecting the stateroot of the trie.

ScottyPoi:
Most recent commits relate to Database research and State Root Indexing research.

- TrieLevel
  - TrieLevel is an implementation of the EthJS `DB` interface that uses an `AbstractLevelDB` like our other History database.
  - EthJS `Trie` class can be created with `Trie.create({ trie: TrieLevel })`
- StateDB
  - StateDB is like a DBManager.
    - sublevels
      - a mapping of state_root -> TrieLevel
    - knownAddresses
      - a `Set` containing all account addresses known to the DB
    - StateRootIndex
      - A custom `graph search` interface 
      - Can sort and find valid orders of known state_roots
 
- StateRootIndex
  - builds a DAG (directed acyclic graph) from sparse account trie data
    - compares changes in nonce / balance per account across stored tries
    - DAG updated with new data with each content update
  - Finds  the longest valid sequences of state_roots
  - Finds the longest valid sequences containing specific state_roots
  - Finds all Possible valid sequences

ScottyPoi (approved):
Merging as scaffolding.

Work continued from here in #406

Issue #408, opened 2023-03-23
Decouple modules from HistoryNetwork types


ScottyPoi:
`Base Protocol` and much of the `Wire` module needs to be decoupled from the history network in a few ways.
Certain things were designed based on assumptions that are no longer valid with the new content types.
ProtocolId.HistoryNetwork is passed as a necessary parameter in methods that should be agnostic of the protocol that called them.  

May require a general refactor of wire spec to better generalize things.

- Types should be made generic, extended by sub-protocols
- Methods should be generalized.
- wire / utp Modules should be agnostic of Protocol specifics

acolytec3:
A lot of this is tackled in #440 (at least with regard to uTP and the `BaseProtocol`).  There's probably some very crafty Typescript ways of making the content types navigable by protocol ID without introducing giant switch statements in every place we have to multiplex between subprotocols but I haven't figured it out yet.

acolytec3:
Going to close this.  I've seen what happens when you try to make giant namespace types to handle stuff and let's just leave it at `switches`.

Issue #406, opened 2023-03-23
State Network Development


ScottyPoi:
- [x] State Network Scaffolding #402 
- [ ] Database Refactor #407 
- [x] JSON RPC #409 #489
- [x] Wire Spec Refactor #408 
- [x] uTP Refactor #408 
- [x] StateProtocol methods #489 
  - [x] Store / Retrieve #407
  - [x] send/handle FindContent
  - [x] send/handle OFFER
  - [x] Gossip
- [ ] Unit testing
- [ ] Integration Test
- [x] CLI devnet #489

acolytec3:
Lets close this and start fresh.

PR #413, opened 2023-03-23
Wire-spec-decoupling

ScottyPoi:
Part of #406 
Addresses #408 

Adds generic typing to BaseProtocol classes.

Adds ProtocolId as an attribute of uTP streams / content lookup / node lookups

Updates History content based tests to explicitly declare protocol id


acolytec3 (approved):
LGTM

PR #420, opened 2023-05-01
Update `history_sendOffer` to match RPC specs

acolytec3:
This revises the implementation of `history_sendOffer` to match the JSON-RPC specs.  All of the `portal-hive` tests specific to the `sendOffer` flow should now pass (at least for `ultralight`/ultralight tests)


ScottyPoi (approved):
LGTM!

I felt torn about the `store` part of it, and whether to enable sending a piece of content just from the RPC request or not.  But I like that we didn't have to make any real changes to the client.

PR #422, opened 2023-07-12
Fix Hive Test Failures

ScottyPoi:
Edits RPC methods to pass hive tests.

Fixes methods to add/get/delete ENR for routing table via RPC

Ultralight passes all current hive rpc-compat tests with this PR.


acolytec3:
Are you opening a separate PR for this one?  I'm happy to review if so.

PR #425, opened 2023-07-23
Revert rpc changes

ScottyPoi:
Reverting some RPC changes from last PR.

The reference tests had been altered by `Trin` team while debugging, and edits that were made to Ultralight mistakenly adjusted to those tests.  This should correct all of the `rpc-compat` tests that broke when the test was restored.


acolytec3 (approved):
LGTM

PR #427, opened 2023-08-01
Adds beacon network content key types

acolytec3:
Adds typing for the beacon network content keys and tests to deserialize the test vectors.


acolytec3:
Admin merging this one (should be noncontroversial and needed for #428 )

PR #428, opened 2023-08-01
Adds handling for content in `beacon network`

acolytec3:
Adds storing and retrieving of `Beacon` network content from local storage as well as a partial implementation of the SENDFINDCONTENT flow.  This currently only handles the case where the content is delivered via uTP stream.


ScottyPoi (approved):
Overall, LGTM!

I think it's worth Typescript engineering something around the content types as we start to add more networks, but that's more about style than function.

:+1 to merge this and then work on setting up new tests

PR #429, opened 2023-08-05
Portal-Hive client-interop fixes

ScottyPoi:
This PR will bring the portal hive client interop tests to ~~100%~~ 98% passing.

* new ENR's are added to routing table by general packet handler
  * Previously this was only available in the SYN packet handling.
* `findENR` functions now use `getWithPending`, which will return the peerId of peers still in the discv5 `pending` buffer. 
* Corrections made to uTP sockets: 
  * `sndId` / `rcvId` / `connectionId` setup
  * `seqNr` / `ackNr` setup
  * `FIN` handling.
    * Corrected to await end of stream before compiling.
* Added debug logging throughout.

### DIFF
--- a/packages/cli/src/rpc/modules/portal.ts
+++ b/packages/cli/src/rpc/modules/portal.ts
@@ -351,13 +351,14 @@ export class portal {
     this.logger(`historyRecursiveFindNodes request returned ${res}`)
     return res ?? ''
   }
-  async historyLocalContent(params: [string]): Promise<string> {
+  async historyLocalContent(params: [string]): Promise<string | undefined> {
     const [contentKey] = params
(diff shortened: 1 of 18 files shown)


ScottyPoi:
Currently `1/73` tests failing:  `OFFER: Receipt  trin --> ultralight`

Should be fixable on Ultralight side.  The failure has to do with `FIN` packet handling, and the fact that `Trin` uTP data packets tend to arrive in an out of order pattern, with the FIN arriving before the final 2 data packets.  

Ultralight's uTP should know to wait for all data packets before compiling the stream, but the current solution for that is insufficient.  

Personally in favor of merging this PR to update the public Portal-Hive tests.  The fix for above can be done in a new PR.

acolytec3 (approved):
LGTM

PR #433, opened 2023-08-09
Update EthJS dependencies to new breaking releases

acolytec3:
Updates all of the EthJS deps to the newest release, also updates discv5 to v5.0.1, a linter dep, and also generally tries to move everything away from buffer as much as possible.

And...it also disables the browser client until we decide what to do with the waku stuff.  Since that's all kinds of broke right now.


acolytec3:
I found a missing dep declaration in `discv5` but nothing directly related to EthereumJS at the moment.

ScottyPoi (approved):
force-push fight!!

LGTM!

Reverted the last commit re: `getSocketAddress` due to import issues.
merging now

acolytec3:
No. 😬 The `buffer` to `uint8array` conversation was the biggest part, and some of the buffer to bytes helpers.

PR #439, opened 2023-08-15
Migrate tests to Vitest

acolytec3:
Migrates tests to use `vitest`

Needs much help


acolytec3:
Commits all got accidentally included in #440 so will close here.

PR #442, opened 2023-08-21
Hive-interop-updates

ScottyPoi:
This PR updates portal network RPC methods to align with spec and pass current hive test suite.

Return type for content lookup changed to `{ content: string, utpTransfer: boolean }`, the expected return type for content related RPC methods.

* Also includes patch for `sendFindNodes` error.


acolytec3 (approved):
LGTM

PR #444, opened 2023-08-28
Add initial support for `LightClientUpdatesByRange`

acolytec3:
This adds a minimal support for the `LightClientUpdatesByRange` in the Beacon network.  

1) Adds an additional `LightClientUpdate` type to the Beacon Network that allows us to store individual light client updates
2) Should seamlessly allow for construction of light client update ranges and then transmit them over the network
3) Removes ENRs from `handleFindContent` message handler for Beacon network since all nodes are expected to maintain all the data for this network


ScottyPoi (approved):
LGTM!

PR #446, opened 2023-08-29
fix: Fixes #445 with functions to check if bootnode and bootnodeList CLI args start with enr and add bootnodes to History Network protocol instance

### LINKED ISSUE #445: should check if each bootnode provide in CLI argument `bootnode` and each `bootnodeList` starts with 'enr:-'
When it is processing the CLI arguments it creates a configured instance of the Portal Network and then uses that to create an instance of the History Network Protocol. It then tries to add the CLI arguments for `bootnode` and `bootnodeList` to the History Network Protocol, but currently it only checks if each bootnode in the `bootnodeList` argument starts with 'enr:-' [here](https://github.com/ethereumjs/ultralight/blob/master/packages/cli/src/index.ts#L145C26-L145C33), but doesn't check if the bootnode provided for the `bootnode` argument starts with 'enr:-' too [here](https://github.com/ethereumjs/ultralight/blob/master/packages/cli/src/index.ts#L139).

ltfschoen:


packages/cli/src/index.ts
```diff
@@ -76,6 +77,19 @@ const reportMetrics = async (req: http.IncomingMessage, res: http.ServerResponse
   res.end(await register.metrics())
 }
 
+const hasBootnodeENRPrefix = (enr) => {
+  if (enr && enr.startsWith('enr:-')) {
```
ltfschoen, line 81: do all the bootnodes (i.e. `bootnode` CLI argument value and all `bootnodeList` CLI argument values) ENR prefixes start with `enr:-`? they do according to the example shown here https://eips.ethereum.org/EIPS/eip-778

but why doesn't the `enr()` validator here https://github.com/ethereumjs/ultralight/blob/master/packages/cli/src/rpc/validators.ts#L227 only check that it starts with `enr` instead of `enr:-`? shouldn't that be updated too for consistency?

packages/cli/src/index.ts
```diff
@@ -76,6 +77,19 @@ const reportMetrics = async (req: http.IncomingMessage, res: http.ServerResponse
   res.end(await register.metrics())
 }
 
+const hasBootnodeENRPrefix = (enr) => {
```
ltfschoen, line 80: could we add import the following
```
import { Enr } from './rpc/schema/types.js'
```
and then explicitly mention the parameter type for the following function and do the same for the `addBootNode` function?
```
const hasBootnodeENRPrefix = (enr: Enr) => {
```

packages/cli/src/index.ts
```diff
@@ -76,6 +77,19 @@ const reportMetrics = async (req: http.IncomingMessage, res: http.ServerResponse
...
+  if (enr && enr.startsWith('enr:-')) {
+    return true
+  }
+  return false
+}
+
+const addBootNode = (protocol, enr) => {
```
ltfschoen, line 87: could we add import the following
```
import { Enr } from './rpc/schema/types.js'
```
and also change this line https://github.com/ethereumjs/ultralight/blob/master/packages/cli/src/index.ts#L2 to the following to import `BaseProtocol` 
```
import { BaseProtocol, PortalNetwork, ProtocolId, fromHexString } from 'portalnetwork'
```
and then explicitly mention the parameter types for the following function?
```
const addBootNode = (protocol: BaseProtocol, enr: Enr) => {
```

it says the type for `protocol` would be `BaseProtocol` here
https://github.com/ethereumjs/ultralight/blob/master/packages/portalnetwork/src/client/client.ts#L36

packages/cli/src/index.ts
acolytec3: I would move these functions to a separate `util.ts` file in the same folder level as this `index.ts` and then add your imports there.

acolytec3: You also need to add appropriate types for each parameter here and in the `addBootNode` helper.

packages/cli/src/index.ts
acolytec3: Yes, bootnode ENRs should all start with `enr:-`.

Feel free to update the `enr()` validator for consistency and add to this PR.

packages/cli/src/index.ts
acolytec3: This is fine, though as noted above, this helper should be exported in a separate `util.ts` and then imported for use here.

packages/cli/src/index.ts
```diff
@@ -135,17 +149,15 @@ const main = async () => {
 
   // TODO - make this more intelligent
   const protocol = portal.protocols.get(ProtocolId.HistoryNetwork)
```
acolytec3, line 151: An ideal solution would be to merge any ENRs provided by `args.bootnode` and `args.bootnodeList`, verify they are valid ENRs, and then loop over the protocols and add each bootnode to each protocol.  Something like:

```js
let validBootnodes = []
if (args.bootnode && isValidEnr(args.bootnode)) validBootnodes.push(args.bootnode)
if (args.bootnodeList) {
    const bootnodeData = fs.readFileSync(args.bootnodeList, 'utf-8')
    const bootnodes = bootnodeData.split('\n')
    for (const bootnode of bootnodes) {
        if (isValidEnr(bootnode)) validBootnodes.push(bootnode)
    }
}

for (const protocol of portal.protocols) {
    for (const bootnode of validBootnodes) {
    addBootNode(protocol, validBootnodes)
    }
}
```

packages/cli/src/index.ts
acolytec3: Also, can you rename this to `isValidEnr` or something similar so the name matches what the function is doing?

packages/cli/src/index.ts
ltfschoen: i did a benchmark comparing `startsWith` and `substr` and found that `startsWith` is about 4-6x faster

```js
function v1(enr) {
  if (enr.substr(0, 3) !== 'enr') {
    return true;
  }
  return false;
}

function v2(enr) {
  if (enr && enr.startsWith('enr:-')) {
    return true
  }
  return false
}

const enr = "enr:-IS4QFV_wTNknw7qiCGAbHf6LxB-xPQCktyrCEZX-b-7PikMOIKkBg-frHRBkfwhI3XaYo_T-HxBYmOOQGNwThkBBHYDgmlkgnY0gmlwhKRc9_OJc2VjcDI1NmsxoQKHPt5CQ0D66ueTtSUqwGjfhscU_LiwS28QvJ0GgJFd-YN1ZHCCE4k"

function runPerformance(...arguments) {
    for (var i=0; i < arguments.length; i++) {
        console.time(arguments[i].name);
        arguments[i](enr);
        console.timeEnd(arguments[i].name);
    }
}

runPerformance(v1, v2);
```

output
```
v1: 0.04ms
v2: 0.009ms
```

ltfschoen: > bootnode ENRs should all start with enr:-.

i understand that they should all start with `enr:`, since it's mentioned in the spec https://eips.ethereum.org/EIPS/eip-778, but i still don't understand why they all have the `-` immediately after `enr:` (i.e. `enr:-`), it seem like the `-` is padding, but i don't understand why they use it at the start and not the end, and why they use `-` instead of `=`. i read that in base64 they use `=` for padding the last character and also sometimes also the second last character.

in this enr library https://github.com/sigp/enr/blob/master/src/lib.rs#L422 they show how they construct the ENR, and they say they construct a _"URL-safe base64 encoded "text" version of the ENR prefixed by "enr:"_

ltfschoen: i was also thinking the `isValidEnr` function that we'll use in util.ts should also verify the signature of the ENR, for example this enr library that's written in Rust allows you to do that here https://github.com/sigp/enr/blob/master/src/lib.rs#L411, but i'm not aware of a JavaScript/TypeScript library that does the same

ltfschoen: i tried checking it with the following with an example bootnode ENR but it doesn't pass the `isValidBase64Length` test
```
function isString(str) {
  return str && typeof str === 'string'
}

function hasValidBase64Chars(str) {
  const validBase64Chars = /[^A-Z0-9+\/=]/i;
  if (str && validBase64Chars.test(str)) {
    return true;
  }
  return false;
}

function isValidBase64Length(str) {
  if (str && str.length && str.length % 4 == 0) {
    return true;
  }
  return false
}

function hasValidEnrPrefix(str) {
  if (str && str.startsWith('enr:-')) {
    return true
  }
  return false
}

function isValidEnr(str) {
  try {
    if (!hasValidEnrPrefix(str)) {
      return false
    }
    // remove initial five characters
    str = str.substr(5, str.length);

    if (!isString(str)) {
      return false
    }
    if (!hasValidBase64Chars(str)) {
      return false
    }
    if (!isValidBase64Length(str)) {
      return false
    }

    return true
  } catch (e) {
    console.error(e)

    return false;
  }
}

const enr = "enr:-IS4QFV_wTNknw7qiCGAbHf6LxB-xPQCktyrCEZX-b-7PikMOIKkBg-frHRBkfwhI3XaYo_T-HxBYmOOQGNwThkBBHYDgmlkgnY0gmlwhKRc9_OJc2VjcDI1NmsxoQKHPt5CQ0D66ueTtSUqwGjfhscU_LiwS28QvJ0GgJFd-YN1ZHCCE4k"

let out = isValidEnr(enr)
console.log(out)
```

acolytec3: So a couple of things:
1) The `enr:` is the 4 character prefix that indicates its a valid ENR.  I didn't realize that at first but reread the ENR spec again they're encoded using Base64URL, not raw Base64 and `-` is a valid character in the Base64URL encoding scheme.  As such, you need to adjust your code to check for valid [Base64url](https://base64.guru/standards/base64url) characters.  
2) The more I think about it, I'm not sure we gain anything by doing all this checking in the top level of our client.  The [ENR][(https://github.com/ChainSafe/discv5/blob/master/src/enr/enr.ts](https://github.com/ChainSafe/discv5/blob/ad1f536480cabf8b0d80916d626c373160b61cc8/src/enr/enr.ts#L165C54-L165C54)) class already checks for the correct prefix before deserializing so I think we're just duplicating checks here.  So, let's redirect this PR and drop all of the ENR validation and instead just merge any supplied ENRs from the two CLI parameters and then maybe wrap the entire loop in a `try/catch` that throws a coherent error regarding invalid ENRs if somebody supplies one.  So, you would just check the error that gets thrown out of the try/catch block,  and then throw a new Error caught by the outer execution loop that explains what the issue is (e.g. `invalid ENR ${badEnr} provided`) or whatever it is.

packages/cli/src/index.ts
ltfschoen: i've made these changes now

ltfschoen:
@acolytec3 i've made the relevant changes, and added some tests for the util.ts functions, but i wasn't successfully able to run the tests. 

i didn't push changes where i tried to get the tests to work, but to explain what i tried, i tried adding the following to the packages/cli/package.json in devDependencies
```
"@vitest/ui": "^0.34.1",
...
"vitest": "^0.34.1",
"webpack": "^5.88.2",
"webpack-cli": "^5.1.4"
```
and added command `"test": "npx vitest run test/* -c=./vitest.config.unit.ts",`
then added a .github/workflows/cli-build.yml file similar to the portalnetwork-build.yml file
and added a vitest.config.unit.ts in packages/cli

but when i try running it with the following using Node.js v18.17.1 LTS

```
cd packages/cli
npm install -g webpack webpack-cli
npm install -D webpack webpack-cli
npm install
```

it gets stuck below when i try installing webpack-cli, even though i've already installed it manually
```
> preinstall
> npm run checkNpmVersion


> checkNpmVersion
> bash ./scripts/check-npm-version.sh

npm >=v7 satisfied

> postinstall
> npm run build --workspaces --if-present


> portalnetwork@0.0.1 build
> tsc


> cli@0.0.1 build
> tsc && cp bootnodes.txt ./dist


> build
> tsc


> browser-client@1.0.0 build
> webpack

CLI for webpack must be installed.
  webpack-cli (https://github.com/webpack/webpack-cli)

We will use "npm" to install the CLI via "npm install -D webpack-cli".
Do you want to install 'webpack-cli' (yes/no): yes
Installing 'webpack-cli' (running 'npm install -D webpack-cli')...
(##################) ⠹ reify:core-js: timing reifyNode:node_modules/react-icons Completed in 1864ms
```

and if i click "No" not to install webpack-cli
it crashes
```
You need to install 'webpack-cli' to use webpack via CLI.
You can also install the CLI manually.
npm ERR! Lifecycle script `build` failed with error:
npm ERR! Error: command failed
npm ERR!   in workspace: browser-client@1.0.0
npm ERR!   at location: /Users/luke/code/clones/github/ltfschoen/ultralight/packages/browser-client
npm ERR! code 1
npm ERR! path /Users/luke/code/clones/github/ltfschoen/ultralight
npm ERR! command failed
npm ERR! command sh -c npm run build --workspaces --if-present

npm ERR! A complete log of this run can be found in: /Users/luke/.npm/_logs/2023-08-31T05_46_10_572Z-debug-0.log
```

the idea was to run the following if that worked
```
npm run test
```

i'll try running `npm install` again and wait a lot longer after saying yes to install webpack-cli again...

ltfschoen:
> @acolytec3 i've made the relevant changes, and added some tests for the util.ts functions, but i wasn't successfully able to run the tests.
> 
> i didn't push changes where i tried to get the tests to work, but to explain what i tried, i tried adding the following to the packages/cli/package.json in devDependencies
> 
> ```
> "@vitest/ui": "^0.34.1",
> ...
> "vitest": "^0.34.1",
> "webpack": "^5.88.2",
> "webpack-cli": "^5.1.4"
> ```
> 
> and added command `"test": "npx vitest run test/* -c=./vitest.config.unit.ts",` then added a .github/workflows/cli-build.yml file similar to the portalnetwork-build.yml file and added a vitest.config.unit.ts in packages/cli
> 
> but when i try running it with the following using Node.js v18.17.1 LTS
> 
> ```
> cd packages/cli
> npm install -D webpack webpack-cli
> npm install
> ```
> 
> it gets stuck below when i try installing webpack-cli, even though i've already installed it manually
> 
> ```
> > preinstall
> > npm run checkNpmVersion
> 
> 
> > checkNpmVersion
> > bash ./scripts/check-npm-version.sh
> 
> npm >=v7 satisfied
> 
> > postinstall
> > npm run build --workspaces --if-present
> 
> 
> > portalnetwork@0.0.1 build
> > tsc
> 
> 
> > cli@0.0.1 build
> > tsc && cp bootnodes.txt ./dist
> 
> 
> > build
> > tsc
> 
> 
> > browser-client@1.0.0 build
> > webpack
> 
> CLI for webpack must be installed.
>   webpack-cli (https://github.com/webpack/webpack-cli)
> 
> We will use "npm" to install the CLI via "npm install -D webpack-cli".
> Do you want to install 'webpack-cli' (yes/no): yes
> Installing 'webpack-cli' (running 'npm install -D webpack-cli')...
> (##################) ⠹ reify:core-js: timing reifyNode:node_modules/react-icons Completed in 1864ms
> ```
> 
> and if i click "No" not to install webpack-cli it crashes
> 
> ```
> You need to install 'webpack-cli' to use webpack via CLI.
> You can also install the CLI manually.
> npm ERR! Lifecycle script `build` failed with error:
> npm ERR! Error: command failed
> npm ERR!   in workspace: browser-client@1.0.0
> npm ERR!   at location: /Users/luke/code/clones/github/ltfschoen/ultralight/packages/browser-client
> npm ERR! code 1
> npm ERR! path /Users/luke/code/clones/github/ltfschoen/ultralight
> npm ERR! command failed
> npm ERR! command sh -c npm run build --workspaces --if-present
> 
> npm ERR! A complete log of this run can be found in: /Users/luke/.npm/_logs/2023-08-31T05_46_10_572Z-debug-0.log
> ```
> 
> the idea was to run the following if that worked
> 
> ```
> npm run test
> ```
> 
> i'll try running `npm install` again and wait a lot longer after saying yes to install webpack-cli again...
> 
> i got the same error downgrading to npm 8.6.0

i had a look at this again and it was because running `npm install` triggers running script `"postinstall": "npm run build --workspaces --if-present",` in project root package.son https://github.com/ethereumjs/ultralight/blob/master/package.json#L18 when its finished and that was causing the error, so i just removed that line temporarily and then `npm install` worked fine, then ran `npm run build --workspace cli --workspace portalnetwork --workspace proxy` and that worked fine, but running `npm run build --workspace browser-client` triggered that error.  anyway, for this PR i don't need to use the browser-client package, but i can't see that error as an open issue yet, should it be added?

ltfschoen:
so i got the tests to work. i added a troubleshooting guide that shows the issues i encountered and how i resolved them and how i got it to work so i could run the CLI package tests on macOS here https://github.com/ethereumjs/ultralight/pull/446/commits/0bb9055f6ddc0d5cae8030746a522118c67381e1#diff-15986190ef9581ab59bcd5483b2c09e7fd0bd439d6f6cddbc94b0b1de094ee51R200

acolytec3:
> @acolytec3 so i got the tests to work. i added a troubleshooting guide that shows the issues i encountered and how i resolved them and how i got it to work so i could run the CLI package tests on macOS here [0bb9055#diff-15986190ef9581ab59bcd5483b2c09e7fd0bd439d6f6cddbc94b0b1de094ee51R200](https://github.com/ethereumjs/ultralight/commit/0bb9055f6ddc0d5cae8030746a522118c67381e1#diff-15986190ef9581ab59bcd5483b2c09e7fd0bd439d6f6cddbc94b0b1de094ee51R200)

Thanks for the hardwork on this.  A couple of further comments:
1) I'm running Ubuntu 22 on an Intel based laptop and I didn't have any of the challenges you noted installing the new deps you added so those may be MacOS specific issues.
2) The only new dependency you should need to run the tests you added is `vitest`.  The `typest/vitest-ui` and the two webpack dependencies aren't necessary.  I removed those 3 dev dependencies, ran `npm run clean && npm i` from the repo root, and it installed everything cleanly.  I was then able to execute `npx vitest run` from the `packages/cli` directory and it ran your tests just fine.  Can you back out those additional deps?

ltfschoen:
> > @acolytec3 so i got the tests to work. i added a troubleshooting guide that shows the issues i encountered and how i resolved them and how i got it to work so i could run the CLI package tests on macOS here [0bb9055#diff-15986190ef9581ab59bcd5483b2c09e7fd0bd439d6f6cddbc94b0b1de094ee51R200](https://github.com/ethereumjs/ultralight/commit/0bb9055f6ddc0d5cae8030746a522118c67381e1#diff-15986190ef9581ab59bcd5483b2c09e7fd0bd439d6f6cddbc94b0b1de094ee51R200)
> 
> Thanks for the hardwork on this. A couple of further comments:
> 
> 1. I'm running Ubuntu 22 on an Intel based laptop and I didn't have any of the challenges you noted installing the new deps you added so those may be MacOS specific issues.
> 2. The only new dependency you should need to run the tests you added is `vitest`.  The `typest/vitest-ui` and the two webpack dependencies aren't necessary.  I removed those 3 dev dependencies, ran `npm run clean && npm i` from the repo root, and it installed everything cleanly.  I was then able to execute `npx vitest run` from the `packages/cli` directory and it ran your tests just fine.  Can you back out those additional deps?

regarding 1. after running `npm run clean` from monorepo root i didn't have any errors running `npm install` from monorepo root anymore, even if i tried each of NPM 9.6.7, 8.6.0, and 7.24.2.
i was able to run the tests with `npx vitest run` too successfully.
i've pushed changes to reflect those findings.

the only issue i appear to still be having is getting the CI tests to pass.
i've gone through and fixed all the linting errors.
but if i run `npm run lint` from monorepo root i get error:
```bash
error: Parsing error: ESLint was configured to run on `<tsconfigRootDir>/test/util.spec.ts` using `parserOptions.project`: /users/me/ultralight/packages/cli/tsconfig.lint.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file at test/util.spec.ts:

error: Parsing error: ESLint was configured to run on `<tsconfigRootDir>/vitest.config.unit.ts` using `parserOptions.project`: /users/me/ultralight/packages/cli/tsconfig.lint.json
However, that TSConfig does not include this file. Either:
- Change ESLint's list of included files to not include this file
- Change that TSConfig to include this file
- Create a new TSConfig that includes this file and include it in your parserOptions.project
See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file at vitest.config.unit.ts:
```

packages/cli/scripts/seeder.ts
```diff
@@ -62,10 +62,10 @@ const main = async () => {
   const peer3 = Client.http({ port: args.rpcPort + 3 })
   const clients = [ultralight, peer1, peer2, peer3]
   const clientInfo: Record<Clients, TestClient> = {
-    ultralight: { client: ultralight, enr: '', nodeId: '' },
-    peer1: { client: peer1, enr: '', nodeId: '' },
-    peer2: { client: peer2, enr: '', nodeId: '' },
-    peer3: { client: peer3, enr: '', nodeId: '' },
+    ultralight: { client: ultralight, enr: 'enr:-IS4QFV_wTNknw7qiCGAbHf6LxB-xPQCktyrCEZX-b-7PikMOIKkBg-frHRBkfwhI3XaYo_T-HxBYmOOQGNwThkBBHYDgmlkgnY0gmlwhKRc9_OJc2VjcDI1NmsxoQKHPt5CQ0D66ueTtSUqwGjfhscU_LiwS28QvJ0GgJFd-YN1ZHCCE4k', nodeId: '' },
```
acolytec3, line 65: Please revert these changes.  This object is populated in the course of running this script - see lines 130-137 - and this causes the CI to fail.

packages/portalnetwork/docs/modules.md
```diff
@@ -547,7 +547,7 @@ Takes an RLP encoded block as a hex string and adds the block header and block b
 | Name | Type | Description |
 | :------ | :------ | :------ |
 | `rlpHex` | `string` | RLP encoded block as hex string |
-| `blockHash` | `string` | block hash as 0x prefixed hext string |
+| `blockHash` | `string` | block hash as 0x prefixed hex string |
```
acolytec3, line 550: These doc files are autogenerated.  Better to fix the spelling in the original typedoc associated with this function and it will get revised the next time we regenerate the docs.

packages/cli/src/rpc/validators.ts
```diff
@@ -224,10 +225,12 @@ export const validators = {
           message: `invalid argument ${index}: enr must be string`,
         }
       }
-      if (params[index].substr(0, 3) !== 'enr') {
+      if (isValidEnr(params[index])) {
         return {
           code: INVALID_PARAMS,
-          message: `invalid argument ${index}: string must start with 'enr:'`,
+          message: `invalid argument ${index}: string must be compliant with \
```
acolytec3, line 231: This shouldn't be multiline as this is part of a JSON response object returned by the RPC.

acolytec3 (changes requested):
A few more small things need fixing before we can merge but this is looking good.

ltfschoen:
i've reverted the changes mentioned and made the JSON response single line

packages/cli/scripts/seeder.ts
acolytec3: You had a bug in the updated enr validator in the RPC.  I just pushed a commit to fix.  CI should pass now so will merge once it does

acolytec3 (approved):
LGTM

PR #450, opened 2023-09-03
UTP-protocol-stream

ScottyPoi:
uTP streams end by emitting a "Stream" event with the compiled content.

While using only 1 network, this was sufficient. 

However, if more than one network is active, the individual subprotocols will each be listening for this `Stream` event, and attempt to process / store any content streamed from uTP.

This PR fixes the issue by changing the `Stream` event.  instead of `Stream`, uTP will emit the protocol id as the event name.

Each subprotocol will listen only for uTP events with its own protocol id.

Note that to do this, the uTP listener setup was moved from the abstract `BaseProtocol` class to the constructors of the individual subprotocols, and each additional sub protocol will need to include this in its constructor.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18767,7 +18767,8 @@
         "prettier": "^2.5.1",
         "ts-node": "^10.4.0",
         "tslib": "^2.3.1",
-        "typescript": "^4.9.4"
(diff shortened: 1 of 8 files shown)


acolytec3 (approved):
lgtm

PR #451, opened 2023-09-03
findNodes: remove self enr exclusion

ScottyPoi:
ultralight fails this portal-hive findNodes test:

`node a` is at distance `d` to `node b`
`node a` sends `findNodes` to `node b` for distance `d`
**expected response**: `NODES: [ node_b.enr ]`

ultralight's `handleFindNodes` method was implemented to exclude a node's own `enr` from a `NODES` response.  

this, like other design choices, was made with the assumption that nodes would always choose to `ping/pong` an unknown node before sending another kind of request.  other portal clients were not built with the same assumptions, which became apparent through various **portal-hive** test results.

this fixes the failing portal-hive interop test

### DIFF
--- a/packages/portalnetwork/src/subprotocols/protocol.ts
+++ b/packages/portalnetwork/src/subprotocols/protocol.ts
@@ -282,8 +282,6 @@ export abstract class BaseProtocol extends EventEmitter {
           nodesPayload.enrs.push(this.enr.toENR().encode())
         } else {
           for (const enr of this.routingTable.valuesOfDistance(distance)) {
-            // Exclude ENR from response if it matches the requesting node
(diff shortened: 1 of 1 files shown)


acolytec3:
The design choice here to exclude our own ENR is because the wire spec says to only send your own ENR if the distance 0 is included in the list of distances.

`Request message to get ENR records from the recipient's routing table at the given logarithmic distances. The distance of 0 indicates a request for the recipient's own ENR record.`

I would say hee that the test is incorrect and that other client implementations are against spec. Besides that, it doesn't make sense to me to send our own ENR by default if we're in the requested distance. The other node will already have our ENR as part of the discv5 session establishing process so this is just extra bytes over the wire.

Is there some discussion somewhere on this?

PR #460, opened 2023-09-13
Add bootstrap retrieval mechanism

acolytec3:
Adds a primitive social bootstrap retrieval mechanism for the Beacon network.  The basic idea is as follows:

When a new node joins the network, it may or may not have a a starting trusted checkpoint.  On the premise that it does not, as the node encounters new peers on the network, it requests the most recent 4 `LightClientUpdates`.  Once it received 5 sets of responses, it counts the number of occurrences of each update's finalized header block roots, sorts these roots by number of "votes" (ie how many responses that finalized root occurred in), and then requests a `LightClientBootstrap` from each node in the voting set in descending order of votes.

To run locally, do the following:
1) Run `npx ts-node-esm scripts/devnet.ts --numNodes=10 --networks=beacon` from the `[root]/packages/cli` directory to start 10 Ultralight nodes running the Beacon network.
2) In a separate terminal, run `npx ts-node-esm scripts/bootstrapFinder.ts`.  This script retrieves the most recent 4 `LightClientUpdates` and the `LightClientBootstrap` associated with the finalized header in each of the updates from a Lodestar mainnet node.  It then distributes the LightClientUpdates to all nodes and then randomly distributes the `LightClientBootstrap` to individual nodes.  It then connects all the nodes together using `addBootNode` and then starts piping in `LightClientOptimisticUpdates` to a single node (who hopefully then gossips it to other nodes).  

If all goes well, you should start to see logs like below:
image


acolytec3:
There's still some sort of error related to not properly cleaning up discv5 after the integration tests run so will work on that but the code as is is ready for review.

packages/cli/scripts/bootstrapFinder.ts
```diff
@@ -0,0 +1,84 @@
...
+    const capellaForkDigest = beaconConfig.forkName2ForkDigest(ForkName.capella)
+    const beaconNode = 'https://lodestar-mainnet.chainsafe.io/'
+    const ultralights: jayson.HttpClient[] = []
+    for (let x = 0; x < 10; x++) {
+        ultralights.push(Client.http({ host: '127.0.0.1', port: 8545 + x }))
+    }
+
+
+    console.log('Retrieving bootstrap and updates from Beacon node...')
+    let optimisticUpdate = ssz.capella.LightClientOptimisticUpdate.fromJson((await (await fetch(beaconNode + 'eth/v1/beacon/light_client/optimistic_update')).json()).data)
```
g11tech, line 26: this can become more convinient if you use `@lodestar/api`:
```typescript
      import {getClient} from "@lodestar/api"
      const api = getClient({baseUrl}, {config});
      api.lightclient.getUpdates(...)
      api.lightclient.getOptimisticUpdate(...)
      ```

File: packages/cli/scripts/bootstrapFinder.ts [hunk 2]
```diff
@@ -0,0 +1,84 @@
...
+import { ForkName } from '@lodestar/params'
+import { computeSyncPeriodAtSlot } from '@lodestar/light-client/utils'
+import { concatBytes, hexToBytes } from '@ethereumjs/util'
+
+
+const { Client } = jayson
+
+const main = async () => {
+    const beaconConfig = createBeaconConfig(defaultChainConfig, hexToBytes(genesisData.mainnet.genesisValidatorsRoot))
+    const capellaForkDigest = beaconConfig.forkName2ForkDigest(ForkName.capella)
```
g11tech, line 17: you can make this more fork agnostic by getting `const fork = config.getForkName(slot)`... but you will anyway get the corresponding `version` (fork) in the beacon api responses and you can use that to create forkdigest for pushing data in portal network

packages/cli/scripts/bootstrapFinder.ts
```diff
@@ -0,0 +1,84 @@
...
+            ),
+        ))
+    }
+    const serializedRange = LightClientUpdatesByRange.serialize(range)
+    const rangeKey = getBeaconContentKey(BeaconLightClientNetworkContentType.LightClientUpdatesByRange, LightClientUpdatesByRangeKey.serialize({ startPeriod: BigInt(oldPeriod), count: 6n }))
+    for (let x = 0; x < 4; x++) {
+        const bootstrapSlot = updatesByRange[x].data.finalized_header.beacon.slot
+        const bootstrapRoot = (await (await fetch(beaconNode + `eth/v1/beacon/blocks/${bootstrapSlot}/root`)).json()).data.root
+        const bootstrap = ssz.capella.LightClientBootstrap.fromJson((await (await fetch(beaconNode + `eth/v1/beacon/light_client/bootstrap/${bootstrapRoot}`)).json()).data)
+        await ultralights[Math.floor(Math.random() * 5)].request('portal_beaconStore', [getBeaconContentKey(BeaconLightClientNetworkContentType.LightClientBootstrap, LightClientBootstrapKey.serialize({ blockHash: hexToBytes(bootstrapRoot) })), toHexString(concatBytes(capellaForkDigest, ssz.capella.LightClientBootstrap.serialize(bootstrap)))])
```
g11tech, line 47: idea of randomzing here is to select from a bigger set? this seems to only go uptill node 6 out of i think 10 nodes? just observing that reselection of a node is also possible (if thats intended behavior here)

g11tech (commented):
looks good to me on high level of the flow :+1:

packages/cli/scripts/bootstrapFinder.ts
acolytec3: Yep, basically just to distribute the bootstraps among several nodes to verify that the process of identifying a good bootstrap by each individual client gets a little more randomness.

g11tech (approved):
lgtm

PR #469, opened 2023-09-27
uTP: fix syn id bug

ScottyPoi:
Reverts change from recent PR.
uTP packet ID numbers were altered in response to hive test failures.
This turned out to be a bug in trin and not actually in ultralight.
This reverts uTP back to spec.


ScottyPoi:
Merging so that public hive tests run with update tonight

PR #470, opened 2023-09-27
uTP: SYN-ACK fix

ScottyPoi:
READ sockets will only receive STATE packets in the form of SYN-ACK during a FINDCONTENT request.
This updates the handler to only expect STATE packet while in connection state: SYN-SENT, rather than comparing the sequence number and ack number.


ScottyPoi:
Tested in local-hive.  merging now for public hive-test consumption.

PR #476, opened 2023-10-06
Browser client updates

acolytec3:
This PR is a potentially temporary measure to speed up our dev workflow
- Renames `client` package to `ui` (we probably could use a better name but `client` is too close to `cli`)
- Changes the name of the `build` script in `browser-client` and `ui` to something other than build so we don't rebuild the browser clients every single time we make a change (which speeds up installs/builds generally ).  When we start working closely on `browser-client` again, it might make sense to change this back.


ScottyPoi (approved):
LGTM!

good call with the 'build' scripts
:up: dev ex

PR #479, opened 2023-10-11
All the things

acolytec3:
This PR does a variety of things
- Adds the ability to report the node type in the `shortId` helper so we can tell if it's trin, ultralight, or fluffy that we're talking to
- Adds some cleanup for the `devnet` script
- Fixes a couple of uTP bugs related to congestion control


ScottyPoi (approved):
LGTM.
gave us improved portal-interop test results. :smile:

PR #481, opened 2023-10-14
UI: Implement Procedures and Interface

ScottyPoi:
# PortalNetwork Browser UI
### PR Changes: 

**ALL tRPC CODE MOVED TO `packages/ui/src/server` **

1. Implements trpc "procedures" to link the UI to a Portal Client.
2. Develops UI to provide a view into Node info, routing table info, bootnode responses, content storage, and message logs.
3. Develops interface to store and retrieve database content from browser.
4. Develops interface to directly input parameters and make JSON-RPC calls in browser, and observe / interact with the results.
5. Debugs technical issues with Websocket and tRPC Servers.
6. Develops App state management
7. Improves App layout and design
8. Adds README.md with quickstart instructions

**packages/portalnetwork/client/**
1. Adds event types for `sentTalkReq` and `sentTalkResp`
2. Adds `logEvents` option to `PortalNetwork` constructor / create function
3. `logEvents` option set to true will enable logs to the UI subscriptions (websockets)


packages/cli/src/ui/server.ts
```diff
@@ -0,0 +1,187 @@
...
+  process.on('SIGTERM', () => {
+    console.log('SIGTERM')
+    handler.broadcastReconnectNotification()
+    wss.close()
+  })
+  portal.discv5.enableLogs()
+
+  portal.enableLog('*ultralight*, *LightClient*, -OFFER, -ACCEPT, *ultralight:RPC*')
+
+  await portal.start()
```
acolytec3, line 169: Just to make sure I understand what this is doing, this is just starting an Ultralight client and relaying the messages over WSS to the UI, right?

acolytec3 (commented):
This looks great as far as it goes.  Should we consider moving the tRPC server instance over to the UI package so we have a clean separation of concerns between the `cli` true NodeJS client and then the one specifically designed to be paired with the UI?

packages/cli/src/ui/server.ts
ScottyPoi: Yes -- this part starts an Ultralight client inside of the server, where we can hook directly into the typescript, and utilize event listeners.

This is unique to the "HTTP" client mode, which communicates with a CLI node running independently

ScottyPoi: Ultimately i should probably decouple this, so that running the internal Ultralight instance is optional, or an entirely different script.

ScottyPoi:
> This looks great as far as it goes. Should we consider moving the tRPC server instance over to the UI package so we have a clean separation of concerns between the `cli` true NodeJS client and then the one specifically designed to be paired with the UI?

:+1:

ScottyPoi:
@acolytec3 

I moved all of the UI related stuff into the UI package, and everything works fine.
I also made the websocket ultralight part default to **off**.  it can be started manually in the UI instead.

This is in a good place to merge now.  Some small code additions into the portalnetwork client code as described above, but everything else is contained to the UI package.

acolytec3 (approved):
Taking it on faith.

PR #488, opened 2023-10-25
State-bridge

ScottyPoi:
Bridge script to generage state network content

from `packages/cli`
run `npx ts-node --esm scripts/bridgeThread.ts`

This will request new block data every 12 seconds using a (free) Alchemy API_KEY, then spin up a worker_thread to request any additional data (proofs, etc.) and generate `state_network` content.

Without any additional options, the script will simply log statistics about the content.

Pass in cli options to pump data into a devnet
`--devnet=true`
`--numNodes`
`--port`

the "port" options refers to the RPC port number of the first node, and assumes that a devnet will consist of nodes at consecutive ports incrementing from this.  It defaults to 8545.

If `numNodes = 1`, the StateBridge script will call `portal_stateStore` to store each new piece of content in the node.

if `numNodes = 2`, the StateBridge script will call `portal_stateStore` to store each new piece of content in the FIRST node.

if `numNodes > 2`, the StateBridge script will call `portal_stateGossip` for each new piece of content, alternating which node is being called with each new block.  This would simulate a network of "bootnodes" each producing a share  of content to pump into the network.


acolytec3 (commented):
Will review this more throughout today.  But one comment is `stateBridge.ts` should be in `cli/scripts` and not part of our source code for the `cli`, right?

ScottyPoi:
> Will review this more throughout today. But one comment is `stateBridge.ts` should be in `cli/scripts` and not part of our source code for the `cli`, right?

Sure.  I didn't really think of this as something to merge the way it is anyway.  i just had it in `cli/src` so that it would automatically compile with `npm run build`

acolytec3 (approved):
LGTM

Issue #490, opened 2023-10-26
Generally improve logging


acolytec3:
Todos for cleaning up logging
- [ ] Put all logging behind a debug flag so it only prints if an environmental variable `DEBUG` is set
- [ ] Have the `enableLog` calls in `CLI` (and wherever else applicable) check for an environmental DEBUG variable for setting topics
- [ ] Update the `shortId` handler to accept an optional routing table variable along with the nodeId so it can lookup the ENR without having to always retrieve an ENR
- [ ] Update all calls to shortId to ensure we add the `c` key if it exists (so we can identify the type of node we're interacting with)
- [ ] Review all of the logging topics and see if we can add more descriptive logging about various protocol behavior

acolytec3:
Obsolete at this point.

PR #492, opened 2023-10-26
Logging cleanup

acolytec3:
Begins work on #490 
- Adds `routingTable` as optional parameter to `shortId`
- Updates all current calls to `shortId` to use the routing table where applicable
- Adds some additional logging in `BeaconLightClientNetwork` when doing different things


ScottyPoi (approved):
lgtm :+1:

PR #499, opened 2023-11-01
chore: update Dockerfile to create 3 layers instead of 10

KolbyML:
image
How it was before ^

image
How it was after ^

I am trying to extract ultralight from the docker image to use in portal-testground a testing framework we are trying out. Currently the ultralight docker image has 10 layers which makes it fairly difficult to get the files I want. Consolidating the layers will make this a lot easier


acolytec3 (approved):
Lgtm

acolytec3 (approved):

PR #501, opened 2023-11-02
State content

ScottyPoi:
Adds StateDB with methods to store and retrieve state content

### DIFF
--- a/packages/portalnetwork/src/subprotocols/state/statedb.ts
+++ b/packages/portalnetwork/src/subprotocols/state/statedb.ts
@@ -39,4 +39,343 @@ export class StateDB {
     this.contractByteCode = new Map()
   }
 
+  /**
+   * Store content by content key
+   * @param contentKey
+   * @param content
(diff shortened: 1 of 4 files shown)


packages/portalnetwork/src/subprotocols/state/statedb.ts
```diff
@@ -0,0 +1,381 @@
...
+import { decodeStateNetworkContentKey } from './util.js'
+
+type StateRoot = string
+type StorageRoot = string
+type CodeHash = string
+type TrieRoot = string
+type Address = string
+
+export class StateDB {
+  trieDB: MapDB<string, string>
```
acolytec3, line 20: Why aren't we using `level` here?

ScottyPoi: we will.  i just sketched it out like this to get it working

acolytec3 (approved):
Lgtm

PR #502, opened 2023-11-02
Upgrade ethers to v6

acolytec3:


packages/portalnetwork/src/client/provider.ts
```diff
@@ -10,69 +10,68 @@ import {
...
   constructor(
-    fallbackProvider: string | ethers.providers.JsonRpcProvider,
+    fallbackProvider: string | ethers.JsonRpcProvider,
     network = 1,
     portal: PortalNetwork,
   ) {
     super(
-      typeof fallbackProvider === 'string' ? fallbackProvider : fallbackProvider.connection,
+      //@ts-ignore
+      typeof fallbackProvider === 'string' ? fallbackProvider : fallbackProvider.#connect.url,
```
ScottyPoi, line 32: what is going on with `fallbackProvider.#connect.url`?

ScottyPoi:
```suggestion
      typeof fallbackProvider === 'string' ? fallbackProvider : fallbackProvider.connection.url,
```

acolytec3: I dunno.  Still trying to work that out. The new and improved `ethers` is quite a bit different than v5.7 was.

ScottyPoi (approved):
LGTM!

PR #507, opened 2023-11-07
portal-client: eth_getBalance

ScottyPoi:
- Adds method to HistoryProtocol to retrieve StateRoot
 - `getStateRoot(blocknumber)`
- Adds method to PortalNetwork for `eth_getBalance`
  - calls `history.getStateRoot` to get Root by block number
  - calls `state.getBalance(address, stateRoot)` to retrieve balance


packages/portalnetwork/src/client/client.ts
```diff
@@ -427,4 +427,23 @@ export class PortalNetwork extends (EventEmitter as { new (): PortalNetworkEvent
       this.emit('SendTalkResp', src.nodeId, requestId.toString(16), toHexString(payload))
     this.discv5.sendTalkResp(src, requestId, payload)
   }
+
+  public ethGetBalance = async (
```
acolytec3, line 431: I think we should move all of these `eth` RPC namespace functions to a dedicated subfolder or set of files instead of spreading them across multiple subprotocols.  Thoughts?

holgerd77: I dreaming for quite some time already of getting the RPC (base) functionality or at least the framing out of client and maybe here as well and unify and then eventually have a wonderfully typed RPC interface definition for the in-s and out-s which is at the same time something like an implicit (explicit?) RPC spec.

Just dropping here, maybe you guys have thoughts on this as well. 🙂

(might be also already something or a level of extradiction, to *just* have an interface definition in the `Util` package which can then the client and Ultralight implement)

ScottyPoi:
Rebasing off PR #509 (post-shanghai blockbody encoding)

Test data is from recent blocks and requires the updated encoding and decoding functions in #509

acolytec3 (approved):
Going to go ahead and approve and merge so we can keep moving on this.

ScottyPoi:
This actually required a bunch of small updates -- which could probably be their own PR's, but :shrug: 
- A few helper functions
- Updates to `addRlpSerializedBlock` and `reassembleBlock`
- Refactors eth_getBlockByNumber to search blockIndex before network
- Refactor eth_getBlockByHash to search local content before network

Now all of our `eth_get` methods will check local content before going to the network.

PR #508, opened 2023-11-07
state: update RPC hooks to use stateDB

ScottyPoi:
Updates `StateProtocol` methods `store` and `findContentLocally` to use `stateDB` methods.

Previously had stored all content directly in DB as key/value pairs.  Now uses StateDB logic to store and serve content

### DIFF
--- a/packages/portalnetwork/src/subprotocols/state/state.ts
+++ b/packages/portalnetwork/src/subprotocols/state/state.ts
@@ -2,7 +2,7 @@ import debug, { Debugger } from 'debug'
 import { PortalNetwork } from '../../client/client.js'
 import { BaseProtocol } from '../protocol.js'
 import { ProtocolId } from '../types.js'
-import { toHexString } from '@chainsafe/ssz'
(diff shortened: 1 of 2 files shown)


packages/portalnetwork/src/subprotocols/state/state.ts
```diff
@@ -158,7 +158,7 @@ export class StateProtocol extends BaseProtocol {
   }
 
   public stateStore = async (contentKey: string, content: string) => {
```
acolytec3, line 160: Why is this called `stateStore` instead of just using the `baseProtocol.store` method below it which is just a noop?

acolytec3 (approved):
LGTM

packages/portalnetwork/src/subprotocols/state/state.ts
acolytec3: Nm, went ahead and renamed things.  I looked at the various usage of `stateStore` and didn't see any reason we needed a separate store method.

PR #509, opened 2023-11-08
BlockBody Encoding

ScottyPoi:
After the addition of withdrawals to the block body in the EIP-4895, clients need to support multiple encodings for the block body content type.

This updates `HistoryNetwork` code with separate types definitions for Pre and Post Shanghai blocks

Also introduces SSZ types for Pre and Post Shanghai block bodies, as well as SSZ encoded `Withdrawals` and SSZ encoded withdrawal arrays (`allWithdrawals`)

`encodeSSZBlockBody` and `decodeSSZBlockBody` were updated to handle both versions.

the `encode` function determines the block type based on the `Block` object input.

`decode`, whose input is an ambiguous Uint8Array, now accepts an optional `withdrawals: boolean = false` input parameter.  By default it will behave as it did previously.  If `withdrawals` is set to `true`, the function will decode the bytes as a post-shanghai block body.  

A `try/catch` is employed as a temporary hack solution to handle cases where `decode` is called without knowing which type of block body to expect.  the function will try to decode as a pre-shanghai blockbody, and then as a post-shanghai blockbody if the deserialize function throws.

For that reason, these changes should be effective and non-breaking without updating any other code.

In practice, a client will always find a header to validate the blockbody, and will know from the header whether the block is pre or post shanghai.


acolytec3 (approved):
LGTM

PR #510, opened 2023-11-08
block index map

ScottyPoi:
## `history.blockIndex: Map<string, string>`

- Adds `blockIndex` as attribute of `HistoryProtocol` class.
- Enables local block lookup by number.

Currently a client cannot perform `eth_getBlockByNumber` without the `EpochAccumulator` that contains that block's `header record`.  

This remains true even if the client already has the block stored locally, because there is no connection between block numbers and block hash without an EpochAccumulator.

`history.blockIndex` is a map of `block number` to `block hash` for content that is in the local DB.  `HistoryProtocol` will record the block number and hash whenever a block is stored.  This way, we can easily retrieve blocks by number from the local database, even if we do not store the relative EpochAccumulator.


acolytec3:
Would it be worth storing this index in the DB so we can reload it on restart?  An in-memory doesn't seem very useful unless we're going to rebuild it on each restart (which in theory could be done by iterating over all the content in the history network db and reconstructing the map.

ScottyPoi:
> Would it be worth storing this index in the DB so we can reload it on restart? An in-memory doesn't seem very useful unless we're going to rebuild it on each restart (which in theory could be done by iterating over all the content in the history network db and reconstructing the map.

good point.  it would also make it universally available for other protocols.  
i'll make that change now

ScottyPoi:
@acolytec3 

Made the suggested change, and got rid of the in-memory map.
Also added a test where a client starts up with a stored block index in its DB

acolytec3 (approved):
LGTM

PR #512, opened 2023-11-09
Make protocols networks

acolytec3:
This is a major house-cleaning PR to bring consistency to our usage of protocol versus subprotocol versus network versus subnetwork.  I checked in the Portal Network discord and it sounds like "Network" is the preferred term for each of our content networks, instead of protocol.  Our codebase has very inconsistent mixing of network and protocol everywhere so this cleans it up.  The only place `protocol` is used is in one `discv5` interface reference since `discv5` uses the term `protocol` to identify what "app-layer" protocol a message is part of when a TALKREQ message is received.


ScottyPoi (approved):
works for me

PR #505, opened 2023-11-06
Enter UltralightStateManager

acolytec3:
Adds an `UltralightStateManager` class that implements the `EVMStateManagerInterface` from `@ethereumjs/statemanager` to allow an evm to retrieve state data from the Portal Network.

PR #515, opened 2023-11-13
Enter `ethCall`

acolytec3:
Adds an initial implementation for `ethCall` to the top level `eth` namespace module.  

This finishes off a feature started in #505 where I manually instantiated an EVM and executed a contract call based on block height and state data stored only in Ultralight.

Note, the current `UltralightStateManager` has no caching and no `put` functionality so only calls that are read only will succeed right now.


ScottyPoi (approved):
Yeah!   This is awesome :smile:

PR #517, opened 2023-11-15
Major Documentation updates

acolytec3:
This PR attempts to drastically revise the documentation for the `cli` and root README docs
- Revises root README to reflect current quickstart usage
- Revises CLI readme by:
  - Moves RPC documentation to new file
  - Cleans up usage guidelines
  - Adds new networks walkthrough that describes how to get a functioning version of each network up and running locally using the various bridge scripts available in `packages/cli/scripts`


ScottyPoi (approved):
Great update

README.md
```diff
@@ -4,47 +4,30 @@ This monorepo comprises a suite of tools **currently in development** to allow d
...
 ## Connecting to the testnet
 
-1.  Follow the above quickstart guide as far as installing all dependencies.
-2.  From the repository root, run `node packages/proxy/dist/index.js --nat=extip`
-3.  Run `npm run start-browser-client`
-4.  Connect to one of the [Ultralight bootnodes](./packages//cli/bootnodes.txt)
-5.  You should start to see the network table populate with additional nodes as the client fills in its routing table
-6.  Try and get any block by blockhash from the Ethereum Mainnet between blocks 1-2500. 
-7.  **Crossing fingers** You should see the block header details before too long
+Follow the above quickstart guide and run `npm run dev-testnet` to start the client with the current list of bootnodes.  The client will attempt to ping all of bootnodes in the provided default bootnode list.
```
holgerd77, line 22: Just running all the instructions.

Will post some observations here post merge.

Dev testnet command seems to work, here some extract from the output:

grafik

It would be nice to have an additional note what kind of network this is actually connecting to (in this case seems: History network) and also to have one addtional sentence on what is then actually happening (just the rough thing ("this is then downloading blocks from PN from mainnet" or whatever is happening).

Also for whole Portal Network, it would be generally good if testnets would have URLs, being it an issue on the Portal network GitHub repo or whatever. Atm these things are so much flurrying around in the sphere ("some" testnet "somewhere" which might or might not persist" 😅).

Doesn't need to be something big, but a few basic infos (when did the testnet start, some few parameters, how long should it run, stuff like that).

holgerd77: 🙂 

Ok, and after one has done this (running the dev command) one is a bit lost about "what to try next". 

Would be nice to have some additional sentence about where one could go from here and what would be a good way to approach. Or whatever makes sense.

holgerd77: Update: ah, just seeing this right only now that you have put all/most of the new documentation in the `cli` folder/package. So that would then be just the missing link I guess, a reference from here to `CLI`, like `See [cli](LINK) package for further instructions on how to connect to various ultralight networks from the CL and information about the networks you can connect to". Or something like that.

README.md
```diff
@@ -4,47 +4,30 @@ This monorepo comprises a suite of tools **currently in development** to allow d
```
holgerd77, line 4: I for my part stumble upon the Ultralight header a bit always.

grafik

Particulary upon the "Clients" in plural and the "in the browser" parts.

The whole things just reads a bit strange.

Maybe "TypeScript" is the more important/graspable/(and: "real" for the moment) part here atm? 🤔

And "Client" instead of "Clients".

So I would suggest: "Portal Network Client in TypeScript"

But you have likely also some thoughts about that.

packages/cli/README.md
```diff
@@ -1,32 +1,31 @@
...
 ## Usage
 
-### Quickstart
-- Clone the ultralight monorepo and run `npm i` from the monorepo root.
-- Change to `./packages/cli`
-- Run `node dist/index.js --proxy --nat=extip --bootnode=enr:[Your favorite bootnode ENR here (if desired)] `
-- The node will start a local instance of the network proxy, determine its public IP address, and then connect to the bootnode specified
+The easiest way to get started with Ultralight is to use the `npm run dev` script to start up a a single local node:
+
+`npm run dev` along with any of the below parameters.
```
holgerd77, line 9: I would personally prefer some command name a bit more expressive here.

Is `ultralight` too much of an "approximation"? I would personally like and this would feasible as a "semantic simplification"? No very strong opinion.

So: `npm run ultralight`

packages/cli/README.md
```diff
@@ -1,32 +1,31 @@
...
+  --bootnodeList      path to a file containing a list of bootnode ENRs      [string]
+  --rpc               Enable the JSON-RPC server with HTTP endpoint          [boolean] [default: true]
+  --rpcPort           HTTP-RPC server listening port                         [number] [default: 8545]
+  --rpcAddr           HTTP-RPC server listening interface address            [string]
+  --metrics           Turn on Prometheus metrics reporting                   [boolean] [default: false]
+  --metricsPort       Port exposed for metrics scraping                      [number] [default: 18545]
+  --dataDir           data directory where content is stored                 [string]
+  --web3              web3 JSON RPC HTTP endpoint for local Ethereum node    [string]
+  --networks          subnetworks to enable  (options are: `history`, `state`, `beacon`) [default: `history`]
+  --trustedBlockRoot  a trusted blockroot to start light client syncing of the beacon chain [string]
```
holgerd77, line 24: A `--help` command would be extremely helpful here, not sure, have you got the infrastructure in place for this that this just drops out?

Then it would be also good if this could be put upon the section here ("Run npm run dev/ultralight -- --help for help").

packages/cli/README.md
```diff
@@ -1,32 +1,31 @@
...
 
-To start a node that has the same node ID each time, you can pass the `--pk` parameter at start-up with a base64 string encoded protobuf serialized private key.  So `ts-node src/index.ts --pk=CAISINx/bjWlmCXTClX2JvDYehb8FSrE6l4MA9LGvP74XdfD` will always start the `cli` client with the node ID `2a9511ca767b7b56bb873234209557d07c5fe09382ed060b272c6a933c5658f5`.
+### Starting with the same Node ID 
 
-You can use the `generateKeys` script to generate any number of private keys using the command `ts-node scripts/generateKeys --numKeys=X` where X is the number of keys to generate.
+To start a node that has the same node ID each time, you can pass the `--pk` parameter at start-up with a base64 string encoded protobuf serialized private key.  So `ts-node-esm src/index.ts --pk=CAISINx/bjWlmCXTClX2JvDYehb8FSrE6l4MA9LGvP74XdfD` will always start the `cli` client with the node ID `2a9511ca767b7b56bb873234209557d07c5fe09382ed060b272c6a933c5658f5`.
 
 ### Connecting to the Portal Network
```
holgerd77, line 30: I think you should really go away from the "to the Portal Network" semantics and move over to always reference a specific network.

holgerd77: (or maybe it's also my lack of the understanding of the PN structure. So are bootnodes e.g. for all the networks together? Or are they always an entrypoint to all the different networks?

(if things like this are e.g. the case then I get the whole `to the Portal Network` notion better, then this might also help though to make correct relation here a bit clearer in the docs))

packages/cli/README.md
```diff
@@ -1,32 +1,31 @@
...
+  --bindAddress       initial IP address and UDP port to bind to             [string]
+  --bootnodeList      path to a file containing a list of bootnode ENRs      [string]
+  --rpc               Enable the JSON-RPC server with HTTP endpoint          [boolean] [default: true]
+  --rpcPort           HTTP-RPC server listening port                         [number] [default: 8545]
+  --rpcAddr           HTTP-RPC server listening interface address            [string]
+  --metrics           Turn on Prometheus metrics reporting                   [boolean] [default: false]
+  --metricsPort       Port exposed for metrics scraping                      [number] [default: 18545]
+  --dataDir           data directory where content is stored                 [string]
+  --web3              web3 JSON RPC HTTP endpoint for local Ethereum node    [string]
+  --networks          subnetworks to enable  (options are: `history`, `state`, `beacon`) [default: `history`]
```
holgerd77, line 23: Bonus suggestion, also 1:1 counting for our client: I think it would be much cooler to have "direct" networks options - since so prominet and important - so `--history`, `--state`, `--beacon`. 🤩 😋

Again, same for EthereumJS client (`--holesky`,...).

packages/cli/README.md
```diff
@@ -39,16 +38,16 @@ Run a local network of CLI Ultralight clients.  Test JSON-RPC calls in a termina
...
 - Run `npm run devnet -- --numNodes=3`
   - This will start 3 nodes with JSON-RPC server addresses `[8545, 8546, 8547]`
-- To specify a port number to, include a `--port` variable
+- To specify a discv5 port number, include a `--port` variable
   - `npm run devnet -- --numNodes=5 --port=9009`
-  - This will start 5 nodes with JSON-RPC server addresses `[9009, 9010, 9011, 9012, 9013]`
+  - This will start 5 nodes with discv5 listener ports on `[9009, 9010, 9011, 9012, 9013]`
 - To specify which subnetworks to support, include one or more `--networks` parameters
-  - `npm run devnet -- --numNodes=5 --port=9009 --networks=history --networks=beacon`
+  - `npm run devnet -- --numNodes=5 --networks=history --networks=beacon`
```
holgerd77, line 45: Is this really correct that this has to be passed by 2x providing the same option? 🤔 So not `--networks=history,beacon`?

packages/cli/README.md
holgerd77: Update: ok, but I understand also if/that you want to make sure that it is clear that these are not the final networks.

holgerd77 (commented):
Ok, couldn't test everything in practice, realized that there is just a bit more work involved (which makes sense), but these are already really great docs, pretty cool! 💯

So especially the three-node-local network instructions, also now the detailed setup for the different (sub)networks. Eager to go deeper her at some point.

README.md
acolytec3: Hehe, definitely, given that we've sort of deprecated the browser client for now anyway.  Will do some further revisions based on the above feedback.  Thanks!

packages/cli/README.md
acolytec3: We use `yargs` in essentially all of these scripts so maybe a generalized "add `--help`" for mroe details on each CLI parameter.

packages/cli/README.md
acolytec3: At present, the bootnodes all just support the History network (and I can't connect to most of them anyway for whatever reason).  That's a big milestone goal for 2024 is to get a more functional "testnet" up and running with responsive nodes that can really handle some load.  The Ultralight bootnodes are all entirely hopeless and basically stop responding within an hour or two of a restart.

packages/cli/README.md
acolytec3: Lemme check if the other option works.  The way I've shown above is how `yargs` processes array params though as far as I know.

PR #519, opened 2023-11-15
Simplify-state-content-types

ScottyPoi:
Both `AccountTrieProof` and `ContractStorageTrieProof` types have been simplified.

The list of witness nodes is all that is necessary.  The extra fields for account or storage data were redundant, as that data is contained in the last witness in the list.

Relevant `state network` code was updated to reflect these changes, as was test data and test scripts.


acolytec3 (approved):
LGTM

PR #524, opened 2023-11-15
Make `eth_call` actually work

acolytec3:
Adds necessary fixes to make `eth_call` work end to end.  

To test: 

Start up Ultralight with state and history activated: `npm run dev -- --networks=state history`

Run the bridge script to add block 3066955 to the state db: `ALCHEMY_API_KEY=[your alchemy API key here] npx ts-node-esm scripts/bridgeThread.ts --devnet --numNodes=1 --block=3066955`

Run `eth_call` as structured below:  This calls a greeter contract at address `0xc54d1a03e5fc1558d4a95e9fbbb2423f02168c23` that reflects your call data back to you - so `0x0000000000000000000000000000000000000000000000000000000000000001` in this case.

`curlie 127.0.0.1:8545 jsonrpc=2.0 id=1 method=eth_call params:='[{ "data":"0x3b421f9e0000000000000000000000000000000000000000000000000000000000000001", "to":"0xc54d1a03e5fc1558d4a95e9fbbb2423f02168c23"}, "0x2ecc4b"]'`


acolytec3:
Note, I also had to change [this check](https://github.com/ethereumjs/ultralight/blob/0a3ab692b237e96122bffb20c09e2d6e871e6b4f/packages/portalnetwork/src/networks/history/history.ts#L134) to only check validate the proof if the block number < 0 (i.e. skip the check)
and similarly [here](https://github.com/ethereumjs/ultralight/blob/0a3ab692b237e96122bffb20c09e2d6e871e6b4f/packages/portalnetwork/src/networks/history/util.ts#L166)

ScottyPoi:
> Note, I also had to change [this check](https://github.com/ethereumjs/ultralight/blob/0a3ab692b237e96122bffb20c09e2d6e871e6b4f/packages/portalnetwork/src/networks/history/history.ts#L134) to only check validate the proof if the block number < 0 (i.e. skip the check) and similarly [here](https://github.com/ethereumjs/ultralight/blob/0a3ab692b237e96122bffb20c09e2d6e871e6b4f/packages/portalnetwork/src/networks/history/util.ts#L166)

this part was reverted, yah?

ScottyPoi:
```
  ddb83:ultralight:RPC Error trying to load block to DB. Error generating inclusion proof: hex argument type undefined must be of type string
```

acolytec3:
> ```
>   ddb83:ultralight:RPC Error trying to load block to DB. Error generating inclusion proof: hex argument type undefined must be of type string
> ```

It worked for me.   I think that's because you need to apply the hack to not check proofs for pre-merge headers.
Let me find a post-merge contract with a pure read function call and we can test it that way without hacks.

acolytec3:
Try this setup instead:
`npx ts-node-esm scripts/bridgeThread.ts --devnet --numNodes=1 --block=18584728`
`curlie 127.0.0.1:8545 jsonrpc=2.0 id=1 method=eth_call params:='[{ "data":"0x95d89b41","to":"0xAb5D6508E4726141D29c6074Ab366aFa03f4Ec8D"}, "0x107aba9"]'`

PR #525, opened 2023-11-16
Fill out `put`/`modify` methods in `UltralightStateManager` and add caches.

acolytec3:
This copies all of the `put`/`modify` namespace methods from the `EthersStateManager` into `UltralightStateManager` along with the necessary caches and tests (also borrowed from `EtherStateManager`)


acolytec3:
Don't merge this until #524 is merged since it's built on top.

PR #532, opened 2023-12-05
Remove restriction on ENRs

ScottyPoi:
- Clients respond to FINDCONTENT requests with an array of ENRs if the target content is not stored locally.

Here, to align with with updates to the portal network wire spec, and other portal clients:

* We remove the conditional in which the client filters the ENR array, only sending nodes **closer to the target than itself**
* Instead Ultralight client will respond an array of ENRs closest to the target content_ID **regardless** of relative distance.

**RPC**

* Also modified the RPC return value for FINDCONTENT.
  * ENRs response updated to return enr strings instead of Uint8Arrays
* Fixes **portal-hive** "portal-mesh" tests


acolytec3 (approved):
Lgtm

PR #535, opened 2023-12-12
state: fix distance error

ScottyPoi:
The distance formula was being used incorrectly in the `state.initGenesis()` process, which caused the wrong set of addresses to be included in the partial state genesis trie.

This fixes it, thus fixing the problem.

This also fixes the test, and the results actually make sense :smile:


acolytec3 (approved):
LGTM

PR #536, opened 2023-12-15
State Network Type Updates

ScottyPoi:
Development branch following changes in PR: https://github.com/ethereum/portal-network-specs/pull/247

Removes Types and Utils for old types
Removes tests and testdata for old types

Removes old logic for address based lookups
Removes old logic for state database 
Removes tests for deleted code.

Disables defunct methods and tests.


ScottyPoi:
@acolytec3 

I narrowed the scope of this down to just implementing the new types and removing the defunct parts of the old implementation.

I'll use the next PR for updates to the state subnetwork class and database

PR #540, opened 2024-02-09
State distance function

ScottyPoi:
Replaces the custom distance function for state network with standard XOR distance function provided in @chainsafe/discv5

State network distance function was originally a modified function creating a "ring geometry" in the DHT.  This has been scrapped for the XOR distance function defined in the wire spec.


ScottyPoi:
Nothing to review here really, just deleting a custom function.

PR #541, opened 2024-02-09
Portal Trie DB

ScottyPoi:
Implements a custom Portal Trie DB class.  PortalTrieDB implements the `DB` interface used in the EthJS Trie contstructor.  This will allow a Trie instance to access the main state network database as well as network retrieval methods during a lookup.

The Trie instance can proceed normally with lookup operations, and use network retrieval methods when nodes are unavailable in the local database.


ScottyPoi:
Merging this with unit tests passing.

Follow-up PR's needed to implement database methods which will utilize this interface.

PR #546, opened 2024-02-19
GetAccount

ScottyPoi:
Implements `getAccount` method in state network

### `getAccount(address, stateroot)`

This method returns an RLP serialized account through a combination of local database, and network requests.

The same way an Execution Client would find an account:

1. Initialize a Trie class
2. Set Trie root to target state root
3. Walk the Trie towards hash of target address
  - Path = `toNibbles(hash(address))`
  - Find current node in DB by hash
  - Decode node to find child hashes
  - Find next node in DB by hash
  - Repeat untill LeafNode is reached
  - Return LeafNode.value()

The only difference occurs when a node is not found in the DB.  In a Portal Client, the unsuccessful database call is repackaged as a network request, and redirected outwards as FINDCONENT.

A node which is obtained this way is not automatically stored in the DB, however, unless it happens to fall in the client's radius.  Instead, nodes are stored in a temporary storage cache, and discarded at the end of the `getAccount` method.

This allows a client to navigate state tries using nodes not stored locally, without also loading the local DB with content outside the client's radius.

### getAccount Test

Included is an **Integration Test** for `state.getAccount`

1. Setup
- A network of 5 PortalNetwork clients are created, each with a radius covering 25% of the keyspace.
- The client's exchange PING/PONG messages, forming a completely connected network.

2. Test Data
- Content for the test is 1 serialized AccountTrieNode OFFER and content key.
- ContentKey includes `node_hash` and `path`
- Content is a 6 node proof for a LeafNode from the genesis Trie.

3. Content distribution
- state.receiveAccountTrieOffer()` is called using one of the clients.
- This method stores in-radius content, and gossips forward an OFFER for the proof of the next node.
- The Gossip process comes to a natural conclusion with gossip of the lone state root.
- The result is each Client in the network storing 2-3 nodes (out of 6 total).

4. Get Account
- A client is chosen who neither stores the state root node, nor the target account leaf node.
- state.getAccount is called using the account `address`
- Client uses Trie methods to navigate the state trie towards the adress.
- Nodes that the client does not store locally are requested as needed from the network using node_hash and the known path to the node in the state trie
- Client getAccount method returns correct account information without storing any more nodes.

This represents the first true Proof-Of-Concept for the updated State Network design.

### DIFF
--- a/packages/portalnetwork/src/networks/state/state.ts
+++ b/packages/portalnetwork/src/networks/state/state.ts
@@ -1,6 +1,6 @@
 import { ENR, distance } from '@chainsafe/discv5'
 import { fromHexString, toHexString } from '@chainsafe/ssz'
-import { BranchNode, Trie, decodeNode } from '@ethereumjs/trie'
+import { BranchNode, ExtensionNode, LeafNode, Trie, decodeNode } from '@ethereumjs/trie'
 import { bytesToHex, bytesToInt, hexToBytes } from '@ethereumjs/util'
 import debug from 'debug'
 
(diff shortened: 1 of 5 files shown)


holgerd77:
Thanks for these excellent PR documentations! 🙏🤩

PR #556, opened 2024-03-27
Add state test vectors to spec test runs

acolytec3:


acolytec3:
No point in keeping this open.

PR #564, opened 2024-05-20
Interop-bugs

ScottyPoi:
Small fixes that address **portal-hive** test failures.

- uTP
  - `FIN` packet handling
  - socket closing
  - handling of `length=0` content
  - log bytes received progress during read stream

- RPC
  - send `PING` during `addEnr` method

- network
  - differentiate content not found and content=`0x` 
  - return `undefined` if not found
  - make `handle` method async

- DB Manager
  - Log during put/get methods

- Discv5
  - tweak timeout settings


acolytec3 (approved):
Just skimmed but lgtm

PR #567, opened 2024-05-21
DB: track currently streaming keys

ScottyPoi:
This adds a list of content keys currently streaming over uTP to the DB Manager.

Before opening an incoming uTP stream, the content key is added to a Set in the DBManager.  

Now, when a request for the content key arrives while the uTP stream is in progress, ultralight will delay the response untill the content key has been successfully added to the DB.  Otherwise, ultralight would respond with 'content not found', because the content was not yet in the DB.  This issue arises in some `portal-hive` tests, which can sometimes proceed faster than the uTP stream, resulting in false negatives.


ScottyPoi:
Despite not changing anything within the state network code -- the changes caused a very bizarre error in the state integration tests.

I investigated this error deeply to find a strange bug.  Somehow, bits of content are getting sliced off and added to other bits of content during gossip.  The result is content arriving in an OFFER which has extra or missing bytes.  Why this is happening is still unknown.  Why this PR exposed this bug is also unknown.  

For now, I wrapped `state.store` in a try/catch so that this error does not hold up what was otherwise passing tests.  The bug should be investigated further, but not in this PR.

PR #569, opened 2024-05-23
Separate bootstrap process from node startup

acolytec3:
This separates out the bootstrap process from node startup so we don't block the main thread while connecting to bootnodes.


ScottyPoi (approved):
LGTM

PR #570, opened 2024-05-23
Various fixes

acolytec3:
- Updates the `devnet.ts` script to use `tsx` so we don't have to recompile the code to see changes we've made to source code in `cli`
- Adds some additional logging in `getBlockByHash`
- Relaxes the requirement to have a header proof when adding a block from an RPC call


ScottyPoi (approved):
LGTM

PR #574, opened 2024-05-28
RPC: wrap PING in try/catch during addENR

ScottyPoi:
The design of `portal-hive` tests causes an error.

`RPC-compat` expects to test `addEnr`, `getEnr`, `deleteEnr` endpoints without actually having an active node with the tested ENR.  So if we try to PING during `addEnr`, we will meet an error.

This PR removes the PING from `addEnr`

This will likely cause errors in other `interop` tests in hive which will need to be addressed separately.


holgerd77:
Just to note: this PR solely deleted a single line (?).

Issue #404, opened 2023-03-14
uTP: Bad format of selective ack extension


kdeme:
I've encountered an error related to https://github.com/ethereumjs/ultralight/issues/403, possibly due to the same underlying issue.

Fluffy logs:

```
WRN 2023-03-14 15:28:33.388+01:00 Failed to decode packet from address       topics="eth utp utp_discv5_protocol" address="(nodeId: dedfdee9cfe1d391f17ad84652cca1e4cdb2d0cb113189dab98564d44bfc31e1, address: 164.92.247.243:5005)" msg="Bad format of selective ack extension: extension = 0 len = 49"
WRN 2023-03-14 15:28:34.573+01:00 Failed to decode packet from address       topics="eth utp utp_discv5_protocol" address="(nodeId: dedfdee9cfe1d391f17ad84652cca1e4cdb2d0cb113189dab98564d44bfc31e1, address: 164.92.247.243:5005)" msg="Bad format of selective ack extension: extension = 0 len = 50"
WRN 2023-03-14 15:28:35.203+01:00 Failed to decode packet from address       topics="eth utp utp_discv5_protocol" address="(nodeId: dedfdee9cfe1d391f17ad84652cca1e4cdb2d0cb113189dab98564d44bfc31e1, address: 164.92.247.243:5005)" msg="Bad format of selective ack extension: extension = 0 len = 54"
WRN 2023-03-14 15:28:35.395+01:00 Failed to decode packet from address       topics="eth utp utp_discv5_protocol" address="(nodeId: dedfdee9cfe1d391f17ad84652cca1e4cdb2d0cb113189dab98564d44bfc31e1, address: 164.92.247.243:5005)" msg="Bad format of selective ack extension: extension = 0 len = 56"
WRN 2023-03-14 15:28:35.618+01:00 Failed to decode packet from address       topics="eth utp utp_discv5_protocol" address="(nodeId: dedfdee9cfe1d391f17ad84652cca1e4cdb2d0cb113189dab98564d44bfc31e1, address: 164.92.247.243:5005)" msg="Bad format of selective ack extension: extension = 0 len = 58"
[...]

PR #575, opened 2024-05-28
Content Lookup improvements

acolytec3:
Applies a variety of fixes/improvements around content lookups

- Adds more descriptive and expansive logging for content lookups
- Adds a `utpTimeout` config option on the `PortalNetwork` client for the uTP streams initiated from the `contentLookup` class.  This should probably be managed lower down in the uTP socket class but it clears out lookups that stall out when a uTP stream fails (likely when Ultralight sends a selective ack packet back and the other node doesn't respond with the in order packets - possibly related to this closed issue - #404 - that we never definitively solved)
- Makes the `eth_getBlockByNumber` test a valid integration test
- Borrows `callWithStackTrace` from the `ethjs` client so we get helpful errors returned from the RPC when calls don't go through correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "eslint-plugin-node": "^11.1.0",
         "eslint-plugin-prettier": "^5.0.0",
         "eslint-plugin-simple-import-sort": "^10.0.0",
+        "node-gyp": "^10.1.0",
         "tsx": "^4.6.2"
       },
       "engines": {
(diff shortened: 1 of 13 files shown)


acolytec3 (commented):
Do not merge yet.  The timeout condition in the contentLookup doesn't seem to be working quite right.

acolytec3:
Gonna keep this ship moving and merge since I don't think there's anything controversial here.

PR #576, opened 2024-05-29
Update Node version for cli package and workflows from version 18 to 20

holgerd77:
This is to break my resilience to do PRs towards and Ultralight repo and maybe it is even doing something useful. 😂

(Node 18 is in maintenance mode https://nodejs.org/en/about/previous-releases and so it is likely time to upgrade, I personally stumbled upon this during installation)

grafik


acolytec3 (approved):
LGTM.  @holgerd77 welcome aboard as a contributor :vulcan_salute:

PR #580, opened 2024-05-31
Remove-libp2p-interface-dep

ScottyPoi:
package: `@libp2p/interface` seems to be a blocker for EthJS Client integration.

This PR removes `@libp2p/interface` as a dependency.  This package was mostly used to declare types for the PeerId and Sepk256 variables during ENR creation, which can be achieved just as easily by declaring those variables as `any`.

We also adjust the client constructor to store the keypair itself instead of the `PeerId`, since accessing the keypair was the only use of the `PeerId` attribute beyond the constructor.


ScottyPoi:
closing as unnecessary for now

PR #582, opened 2024-06-14
Individual Network Config

ScottyPoi:
Clients should be able to configure individual networks with unique `radius` settings.  

However, all `supported subnetworks` are initialized using a global radius value.  

This adds a type: `NetworkConfig`, and updates the type `PortalNetworkOpts` to accept as `supportedNetworks` a list of `NetworkConfig` options instead of simply a list of `NetworkId`.  `NetworkConfig` includes `NetworkId` as well as a `radius` value.

`PortalNetwork` client code was updated to assign each network its unique radius value.

CLI and test scripts were updated to pass the new `NetworkConfig` list into client constructors. 

```
export interface NetworkConfig {
  networkId: NetworkId
  radius: bigint
}

export interface PortalNetworkOpts {
  supportedNetworks?: NetworkConfig[]
  ...
}
```

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -83,6 +83,21 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     array: true,
     optional: true,
   })
+  .option('radiusHistory', {
+    describe: `2^r radius for history network client`,
+    number: true,
+    default: 0,
+  })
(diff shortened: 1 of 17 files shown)


ScottyPoi:
will have to update https://github.com/ethereumjs/ethereumjs-monorepo/pull/3444 if merged

PR #584, opened 2024-06-20
Parallel lookups in recursive find content

ScottyPoi:
This improves the efficiency of recursive content lookups

We now make 5 requests at once to 5 closest peers:
```
while (!this.finished && this.lookupPeers.length > 0) {
      // Process multiple peers in parallel
      const peerBatch = this.lookupPeers.splice(0, 5)
      const promises = peerBatch.map((peer) => this.processPeer(peer))

      const results = await Promise.all(promises)

      for (const result of results) {
        if (result) {
          this.finished = true
          return result
        }
      }
    }
```

A new method: `processPeer` performs the content request and handles the response.

If there is no response, `processPeer` returns undefined.
If the response is an ENR list, the ENR's are added to the lookup, and `processPeer` returns **undefined**

If the reponse leads to a successfull content stream, either directly or via uTP, `processPeer` returns the **content response**

`ContentLookup` will proceed recursively in the same way as before, however will now make 5 parallel requests at a time in the process.  This should lead to more efficient lookups.

### DIFF
--- a/packages/portalnetwork/src/networks/contentLookup.ts
+++ b/packages/portalnetwork/src/networks/contentLookup.ts
@@ -73,9 +73,9 @@ export class ContentLookup {
       const results = await Promise.allSettled(promises)
 
       for (const result of results) {
-        if (result) {
+        if (result.status === 'fulfilled' && result.value !== undefined) {
           this.finished = true
-          return result
+          return result.value
         }
       }
     }
--- a/packages/portalnetwork/src/networks/contentLookup.ts
+++ b/packages/portalnetwork/src/networks/contentLookup.ts
@@ -70,8 +70,7 @@ export class ContentLookup {
       // Process multiple peers in parallel
       const peerBatch = this.lookupPeers.splice(0, 5)
       const promises = peerBatch.map((peer) => this.processPeer(peer))
-
-      const results = await Promise.all(promises)
+      const results = await Promise.allSettled(promises)
 
       for (const result of results) {
         if (result) {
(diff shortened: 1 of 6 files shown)


packages/portalnetwork/src/networks/contentLookup.ts
```diff
@@ -65,121 +66,120 @@ export class ContentLookup {
...
-        this.logger(
-          `No more peers to query.  Failed to retrieve ${toHexString(this.contentKey)} from network`,
-        )
-        return
+    while (!this.finished && this.lookupPeers.length > 0) {
+      // Process multiple peers in parallel
+      const peerBatch = this.lookupPeers.splice(0, 5)
+      const promises = peerBatch.map((peer) => this.processPeer(peer))
+
+      const results = await Promise.all(promises)
```
acolytec3, line 74: I'm not sure I like the idea of `Promise.all` since it will reject if any of the batch of promises rejects (and it looks like we currently reject if the uTP stream times out without getting us the desired content) and I think that will end up crashing the lookup (unless we handle it somehow).

I've had a slight rethink of this and now think something like this might work better (not sure yet so would need to test out)
My rough idea is something like this:`
`while !finished {`
1) Add a new array of `this.lookupPromises`
2) `for (peer of firstFivePeers) void processPeer` 
3) Do `await new Promise((resolve) => setTimeout(3000, () => resolve))` or some suitable timeout
4) See if any of the promises have resolved.  If so, `for (peer of upToFiveNewPeers) void processPeer`
`}`

We'd need to pass a `resolve` function or something to `processPeer` that could call to mark "finished" and then close the lookup.  We also need to make sure that none of the promises reject without exceptions being caught. I think you can get away with `void promise.catch(() => handle exception)`.  Lemme do some experimenting over the next few days and see.

ScottyPoi: Good point.  How about `Promise.allSettled`?  It resolves when all internal promises are resolves or rejected.

acolytec3 (approved):
LGTM. Will follow up with better concurrency if feasible down the line

PR #588, opened 2024-06-27
Reject out of radius

ScottyPoi:
Issue:  client accepts `contentKeys` from `OFFER` with `contentId` outside of the `nodeRadius`  

When the client sends gossip, it only sends to those peers for whom the `contentId` lies in their node radius.

However, upon receiving gossip, the client did not compare the incoming content id's to its own radius, instead, accepting every piece of offered content not already stored.  

This PR adds a distance check to the `handleOffer` method, which filters out content not in the node's radius.  

-- 
**Note** This update will cause some `portal-hive` tests to fail.  This is due to Ultralight defaulting to a radius of 0.  A PR has been submitted to the `hive` repo adding commandline flags to the startup script that set the node radius to `2^256 - 1`.  Merging that update into hive should resolve the broken test cases.

### DIFF
--- a/packages/portalnetwork/src/networks/beacon/beacon.ts
+++ b/packages/portalnetwork/src/networks/beacon/beacon.ts
@@ -1,3 +1,4 @@
+import { digest } from '@chainsafe/as-sha256'
 import { toHexString } from '@chainsafe/ssz'
 import {
   bytesToHex,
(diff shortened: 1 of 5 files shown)


acolytec3 (approved):
LGTM

PR #595, opened 2024-07-03
Configure storage size not radius

ScottyPoi:
PortalNetwork users should not need to configure the node's **radius** directly.
**Radius** refers to the "area" of the network for which the node claims responsibility.  A node "should" store the content "in its radius".

However, what a user will actually care about (and want to configure) is the **storage space** allocated to each network.  It is this setting which should determine the correct radius for a node, not the other way around.  

With these changes, users will set a **maxStorage** size instead of a **radius** value for each subnetwork.  This will default at  `1024 MB`.  The `radius` for each network will *start* at `100%` (or `2^256`) .  As content is added, the node will evaluate the storage used, and prune to a smaller radius as needed to free up storage space.

Content that is pruned (deleted from the DB) will be gossiped back into the network to protect against loss.

TODO:  
1. Decide when and how often to check db size 
  - On an `interval`?
  - During `handleOffer` ?
    - Every time?
    - Random?
2. Determine threshold for pruning
  - `size > maxStorage` ?
  - `size > maxStorage * p%` ?

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -83,20 +83,20 @@ const args: ClientOpts = yargs(hideBin(process.argv))
     array: true,
     optional: true,
   })
-  .option('radiusHistory', {
-    describe: `2^r radius for history network client`,
+  .option('storageHistory', {
+    describe: `Storage space allocated to HistoryNetwork DB in MB`,
     number: true,
-    default: 256,
+    default: 1024,
   })
-  .option('radiusBeacon', {
-    describe: `2^r radius for beacon network client`,
+  .option('storageBeacon', {
+    describe: `Storage space allocated to BeaconChainNetwork DB in MB`,
     number: true,
-    default: 256,
+    default: 1024,
   })
-  .option('radiusState', {
-    describe: `2^r radius for state network client`,
+  .option('storageState', {
+    describe: `Storage space allocated to StateNetwork DB in MB`,
     number: true,
-    default: 256,
+    default: 1024,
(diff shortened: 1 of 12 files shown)


packages/cli/src/index.ts
```diff
@@ -83,20 +83,20 @@ const args: ClientOpts = yargs(hideBin(process.argv))
...
-    default: 256,
+    default: 1024,
   })
-  .option('radiusBeacon', {
-    describe: `2^r radius for beacon network client`,
+  .option('storageBeacon', {
+    describe: `Storage space allocated to BeaconChainNetwork DB in MB`,
     number: true,
-    default: 256,
+    default: 1024,
```
acolytec3, line 94: Beacon Network shouldn't need anywhere near a GB. It's just the bootstrap, a few light client updates, and two headers.

acolytec3 (approved):
LGTM though I would never expect Beacon to hit it's limit.

ScottyPoi:
> LGTM though I would never expect Beacon to hit it's limit.

Yeah I picked 1GB somewhat arbitrarily as a non-trivial yet rather painless default commitment.

holgerd77:
Makes a lot of sense! 🙏

ScottyPoi:
**update**

current strategy:
- prune() called when DB is initialized.
- if `metrics` enabled:
  - prune() called every 50 `OFFER`s received.
- if no `metrics`  
  - prune() called randomly 1/50 `OFFER`s

PR #596, opened 2024-07-05
Add `historicalRoots` accumulator and proof types

acolytec3:
This adds the `historicalRoots` accumulator from the beacon state that represents hashes for historical block headers between the Bellatrix/Merge block and the Capella/Shanghai hardforks and associated ssz types necessary to provide block headers with proofs for this block range.

It also updates the `portal-spec-tests` to the latest commit and adds a type serialization test for a post merge header.

Lastly, removes multiple duplicate versions of the pre-merge accumulator we had in the `portalnetwork/networks/history/data` subdirectory and moves a binary encoded version of it to the `test` directory since we use the JSON object equivalent in Ultralight proper.


acolytec3 (commented):
This builds on the type definitions in [this PR to the portal network specs](https://github.com/ethereum/portal-network-specs/pull/287)

ScottyPoi (approved):
LGTM

PR #599, opened 2024-07-08
Fix DB size scale mismatch

ScottyPoi:
issue: 

`network.prune` calls `this.db.size()`, and compares the result to the `maxStorage` setting.

The problem is that `db.size()` reports the size in **bytes**, and `maxStorage` is a **megabyte** value.  

fix:

`maxStorage` is multiplied by 1,000,000 to match the scale


acolytec3 (approved):
LGTM

PR #607, opened 2024-07-22
uTP: refactor ContentReader (and bug fixes)

ScottyPoi:
### ContentReader

`ContentReader` is a  class object used by a uTP **READ** Socket to organize and compile the bytes streamed in DATA packets.

The class is initialized with a **starting sequence number**, set in the first exchange of SYN/STATE packets.  Subsequent DATA packets are collected and ordered.  Upon learning of the **ending sequence number** from a FIN packet, the content reader will wait for any stray packets to arrive, and then produce the full bytestream.

For FINDCONTENT streams, this process reamins unchanged.  For OFFER streams, in which multiple pieces of content may be streamed together, we now read the bytes as they arrive.  The first bytes of any piece of content in a stream describe the `length` and `offset` of the next piece of data.  Using this information, we are able to determine when a piece of content as fully streamed, even as more data packets continue to arrive, which means we can store content as we go, rather than wait for the entire stream to end.

This provides a small performance and memory gain.  More importantly, thought, it hardens against content loss when streams fail part way.  This way, content that successfully arrived in a stream that ultimately fails for some reason will not be lost.

### Bug Fixes

This work led to some deep investigation of the *random-ish* failures of hive interop tests.

I discovered that the failures occured, usually, when one of the first, and/or last DATA packets arrived out of order.  Out of order packts  are something which uTP is generally designed to be able to handle.  Since the first and final packets in a uTP exchange also inform the receiving socket of the start/end of the data stream, occasionally `Ultralight` would use an OOP packet to set the expecations for sequence numbers.

The fix is a few extra conditional within handling of DATA packets which allow uTP to self correct in these situations.  

**These fixes led to 100% passing rate on `hive` interop tests**


packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
```diff
@@ -271,42 +282,45 @@ export class UtpSocket {
...
       await this.sendAckPacket()
-      return _content
+      this.close()
+      this._clearTimeout()
+      if (compile === true) {
+        return this.finish()
+      }
     } else {
+      this.logger.extend('FIN')(`Expected: ${this.ackNr + 1} got ${packet.header.seqNr}`)
       // TODO: Else wait for all data packets.
```
acolytec3, line 313: Where should this happen?  Should we trigger `handleFinPacket` from `contentReader.addPacket` if an OOO packet is received and this finishes all the data in the packet?

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
```diff
@@ -271,42 +282,45 @@ export class UtpSocket {
...
+      )
       this.seqNr = this.seqNr + 1
       this.ackNr = packet.header.seqNr
-      const _content = await this.reader!.run()
-      this.reader = undefined
-      this.logger(`Packet payloads compiled into ${_content.length} bytes.  Sending FIN-ACK`)
-      this.close()
       await this.sendAckPacket()
-      return _content
+      this.close()
```
acolytec3, line 306: Why not just pass the compile flag to `finish` so we don't have to duplicate this code?

acolytec3 (approved):
Will approve since this is an improvement as is but left a couple of comments we could improve upon.

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
ScottyPoi: This is somewhat non-critical (and an old TODO ).
What is left somewhat un-finished here is that if a FIN arrives before the last DATA packets, we won't end up sending the FIN-ACK packet.  But since we will have ACK'ed all of the DATA packets, the other socket doesn't really need to see that packet

packages/portalnetwork/src/wire/utp/Socket/UtpSocket.ts
ScottyPoi: good call.
adjusted the naming a bit, too.  `finish` is now `compile`, and is called by `close` if `compile = true`.

ScottyPoi:
Addressed your comments and cleaned up the `close` / `finish` methods.
Will merge once hive runs once more

PR #608, opened 2024-07-27
portalnetwork: fix for DB issue

ScottyPoi:
Bootnodes were not pruning DB's correctly.

This fixes ( and tests ) DB bug


ScottyPoi:
merging without review so that i can do bootnode maintenance today

PR #609, opened 2024-07-31
Database: directly use content keys

ScottyPoi:
Database Keys were previously based on `distance` calculation.

This allowed for optimized database pruning by sorting content automatically by distance and using the native `gte` base iterator from level db.

However, it did not allow for re-gossiping content during the pruning process, as there is no direct way to recalculate content KEYS from the database key.

It also meant that a database was only useful for a node with a specific node id.  if a client were to restart with a new nodeID, the database would be unusable.

Solution:

- Store content in the database directly by content key.
- Remove `databaseKey` function
- Refactor `prune` to calculate distance 

--

This PR will be breaking for any active ultralight databases.

### DIFF
--- a/packages/portalnetwork/src/networks/networkDB.ts
+++ b/packages/portalnetwork/src/networks/networkDB.ts
@@ -65,16 +65,13 @@ export class NetworkDB {
   async put(key: string, val: string) {
     if (!key.startsWith('0x')) throw new Error('Key must be 0x prefixed hex string')
     if (!val.startsWith('0x')) throw new Error('Key must be 0x prefixed hex string')
-    const databaseKey = this.databaseKey(key)
     try {
-      await this.db.put(databaseKey, val)
+      await this.db.put(key, val)
     } catch (err: any) {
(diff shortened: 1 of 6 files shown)


packages/portalnetwork/src/networks/networkDB.ts
```diff
@@ -166,9 +149,15 @@ export class NetworkDB {
    * Prune DB content to a new radius
    * @param radius node radius
    */
-  async prune(radius: bigint): Promise<void> {
-    for await (const key of this.db.keys({ gte: bigIntToHex(radius) })) {
-      await this.db.del(key)
+  async prune(radius: bigint): Promise<[string, string][]> {
```
acolytec3, line 152: Should we keep an in memory list of DB keys so we don't have to read from disk just to figure out if a locally stored key can be pruned or not?  I'm thinking just an LRU cache that sits in the networkDB and contains a list of all locally stored keys.  It probably has numerous other performance benefits for down the line (i.e. when we get a FINDCONTENT or OFFER message, we can just check the in memory cache to see if we're storing the content instead of reading from disk)

PR #611, opened 2024-08-07
Store Network radius value in DB

ScottyPoi:
Adds async method `setRadius` to `network` classes.

setRadius sets the `nodeRadius` attribute of the class, but now also stores that value in the network DB (as a string).

during `client.start`, the client will check each network's DB for a stored `radius` value, and call prune.

This way, a client that is restarted with a preexisting DB will immediately remember its own radius settings.  Currently, the radius ends up getting reset to 100%, causing the DB to bloat temporarily before pruning again.  This is likely the cause of the current public bootnode instability.


acolytec3 (approved):
LGTM

PR #613, opened 2024-08-23
Add era file support

acolytec3:
This adds a new package specifically for support for ERA files which store Beacon chain history and state.  This is a set of utilities that we can use generating header proofs for post Capella blocks.


packages/era/src/types.ts
```diff
@@ -0,0 +1,17 @@
+export const EraTypes = {
+  CompressedSignedBeaconBlockType: new Uint8Array([0x01, 0x00]),
+  CompressedBeaconState: new Uint8Array([0x02, 0x00]),
+  Empty: new Uint8Array([0x00, 0x00]),
+  SlotIndex: new Uint8Array([0x69, 0x32]),
+}
+
+export type e2StoreEntry = {
+  type: Uint8Array
```
ScottyPoi, line 9: Will this `type` value be one of the `EraTypes` above?

ScottyPoi (commented):
LGTM.

Couldn't we just include this in the `cli` package?  Or is the idea that you would use this package manually prior to setting up a bridge node?

acolytec3:
Going to go ahead and merge so I can continue on the Historical Summaries PR.

PR #616, opened 2024-08-26
Generate and verify post capella proofs

acolytec3:
Adds support for generating and verifying post-capella proofs rooted in `HistoricalSummaries`.

This also updates the various lodestar deps to v1.21


packages/portalnetwork/src/networks/beacon/types.ts
```diff
@@ -57,7 +57,8 @@ export const HistoricalSummariesStateProof = new VectorCompositeType(Bytes32Type
 export const HistoricalSummariesWithProof = new ContainerType(
   {
     epoch: new UintBigintType(8),
-    historicalSummaries: ssz.allForks.capella.BeaconState.fields.historicalSummaries,
+    // TODO: Figure out if there's a way to make this for "all forks post capella"
```
g11tech, line 60: historialSummaries type is fixed for all forks do your current way to get the container from capella is good

packages/portalnetwork/src/networks/history/util.ts
```diff
@@ -294,13 +295,60 @@ export const verifyPreCapellaHeaderProof = (
     'blockHash',
   ])
   const reconstructedBlock = ssz.bellatrix.BeaconBlock.createFromProof({
```
g11tech, line 297:
```suggestion
  const reconstructedBlock = ssz[fork[.BeaconBlock.createFromProof({
```

?

packages/portalnetwork/src/networks/history/util.ts
```diff
@@ -294,13 +295,60 @@ export const verifyPreCapellaHeaderProof = (
...
   })
 
-  if (!equalsBytes(reconstructedBlock.hashTreeRoot(), proof.beaconBlockHeaderRoot)) return false
+  if (!equalsBytes(reconstructedBlock.hashTreeRoot(), proof.beaconBlockRoot)) return false
+  return true
+}
+
+export const verifyPostCapellaHeaderProof = (
+  proof: ValueOfFields<{
+    beaconBlockProof: VectorCompositeType<ByteVectorType>
```
g11tech, line 310: i think you could simplify this to Uint8Array[]

packages/portalnetwork/src/networks/history/util.ts
```diff
@@ -294,13 +295,60 @@ export const verifyPreCapellaHeaderProof = (
...
+    witnesses: proof.historicalSummariesProof,
+    type: ProofType.single,
+    gindex: historicalSummariesPath.gindex,
+    leaf: proof.beaconBlockRoot, // This should be the leaf value this proof is verifying
+  })
+
+  if (
+    !equalsBytes(
+      reconstructedBatch.hashTreeRoot(),
+      historicalSummaries[Number(slotToHistoricalBatch(proof.slot)) - 758].blockSummaryRoot,
```
g11tech, line 334: probably store 758 in some constant for later clarity/recall

packages/portalnetwork/src/networks/history/util.ts
acolytec3: I'm not sure what you mean though I just realized I have the wrong type here since this is a list now. How would I switch it to a `Uint8Array[]`?

packages/portalnetwork/src/networks/history/util.ts
acolytec3: In this instance, this function should only ever be called for post merge pre Capella blocks, so we can safely stick with the bellatrix types here.

packages/portalnetwork/src/networks/history/util.ts
acolytec3: Good call.  Have done.

PR #622, opened 2024-09-09
History: header lookup by number

ScottyPoi:
Implements changes in this proposal: https://github.com/ethereum/portal-network-specs/pull/334

### Removes EpochAccumulators from `HistoryNetwork`

History Network will no longer store Epoch Accumulators.  

The purpose of these accumulators in the network were:
1. Generating pre-merge header proofs
2. Canonical lookup of blockhash by number

The epoch accumulators can be downloaded and stored locally for these reasons.  They are a static data set.  The pre-merge header proofs have been generated and gossiped and can be recreated from an accumulator if it becomes unavailable in the network.

### Updated History Network methods / functions

1. validateHeader
2. getBlockHeaderFromDB
3. getBlockFromDB
4. getContentKey
5. getContentId
6. decodeHistoryNetworkContentKey
7. prune

### Block Index

Portal Client maintains an index of block number > block hash.   `getBlockHeaderFromDB`, `getBlockFromDB`, and other methods actively use this index to aid in block lookups.

### eth_getBlockBy... methods

1. `eth_getBlockByNumber` will now directly query HistoryNetwork for block header by number
2. `eth_getBlockByHash` will also query network for block header by number if block header is not found by hash, but the block number can be found in the block index

Index of block hash to number can be generated from stored accumulators if desired.

The `0x03` content type prefix will be deprecated for this type and reused for the new type

### BlockHeaderByNumber ContentType

The `0x03` content type prefix will now be used by a BlockHeader_by_Number content key.

The content retrieved with this key is the same `BlockHeaderWithProof` type stored with content key prefix `0x00`

**Content will NOT be stored in the DB under these keys**

`BlockHeaderWithProof` content will still be stored using the `0x00` prefix contentKey derived from the **HASH**.  When `history.store` is called for a `0x03` content key, the content key will be switched to the hash key before storing in the database.

Lookups for `0x03` prefixed keys will reference the **BlockIndex** and redirect to a database lookup for the HASH based key.  This switch occurs in the `history.findContentLocally` method.

```python
block_number_key = Container(block_number: uint64)
selector         = 0x03

block_header_with_proof = BlockHeaderWithProof(header: rlp.encode(header), proof: proof)

content          = SSZ.serialize(block_header_with_proof)
content_key      = selector + SSZ.serialize(block_number_key)
```

## Prune

Special conditions must now be considered during history DB pruning.  The stored (hash based) content key MAY fall outside of the radius while the **number** based content key remains inside the radius.  During pruning, the second contentId must be calculated for any `0x00` prefixed content keys, and content should only be pruned if BOTH keys fall outside the radius.


packages/portalnetwork/src/client/eth.ts
```diff
@@ -1,15 +1,15 @@
 import { fromHexString, toHexString } from '@chainsafe/ssz'
 import { EVM } from '@ethereumjs/evm'
 import { Address, TypeOutput, bytesToHex, hexToBytes, toType } from '@ethereumjs/util'
+import { keccak256 } from 'ethereum-cryptography/keccak.js'
```
holgerd77, line 4: Can you do the imports from Util? Maybe you guys also add this listing rule to disallow ethereum-cryptography imports (bigger project: adjust *all* listing rules)?

holgerd77:
🎉

packages/portalnetwork/src/client/eth.ts
ScottyPoi: import `keccak256` from `@ethereumjs/util`?

packages/portalnetwork/src/client/eth.ts
```diff
@@ -112,6 +119,28 @@ export class ETH {
...
+  getBlockByTag = async (
+    blockTag: 'latest' | 'finalized',
+    includeTransactions: boolean,
+  ): Promise<Block | undefined> => {
+    // Requires beacon light client to be running to get `latest` or `finalized` blocks
+    this.networkCheck([NetworkId.BeaconChainNetwork])
+    let clHeader
+    switch (blockTag) {
+      case 'latest': {
+        clHeader = this.beacon!.lightClient?.getHead() as capella.LightClientHeader
```
acolytec3, line 132: Not to resolve here but note to self to update the ssz typings.  We should be able to make this future proof if I can remember how to do it.

packages/portalnetwork/src/networks/history/util.ts
```diff
@@ -41,42 +42,74 @@ import type {
 import type { WithdrawalBytes } from '@ethereumjs/util'
 import type { ForkConfig } from '@lodestar/config'
 
+export const BlockHeaderByNumberKey = (blockNumber: bigint) => {
+  return Uint8Array.from([0x03, ...BlockNumberKey.serialize({ blockNumber })])
```
acolytec3, line 46: Is `0x03` the ContentType?  If so, can we replace `0x03` with `HistoryNetworkContentType.BlockHeaderByNumber`?

packages/portalnetwork/src/networks/history/util.ts
```diff
@@ -41,42 +42,74 @@ import type {
...
     default:
       throw new Error('unsupported content type')
   }
   return encodedKey
 }
-export const getContentId = (contentType: HistoryNetworkContentType, hash: string) => {
-  const encodedKey = hexToBytes(getContentKey(contentType, hexToBytes(hash)))
-
+export const getContentId = (contentType: HistoryNetworkContentType, key: Uint8Array | bigint) => {
+  const encodedKey = hexToBytes(getContentKey(contentType, key))
```
acolytec3, line 82: Here again, this is a future todo but we should figure out how to minimize the conversions between bytes and strings.  I think we go back and forth a lot needlessly and should stick with one internal representation of hashes/keys.  I'm guessing a main driver here is the db keys, but I'm guessing we could use uint8arrays there if needed.

packages/portalnetwork/src/networks/history/history.ts
```diff
@@ -52,6 +52,19 @@ export class HistoryNetwork extends BaseNetwork {
    * @returns content if available locally
    */
   public findContentLocally = async (contentKey: Uint8Array): Promise<Uint8Array | undefined> => {
+    const contentType = contentKey[0]
+    if (contentType === HistoryNetworkContentType.BlockHeaderByNumber) {
+      const blockNumber = '0x' + decodeHistoryNetworkContentKey(contentKey).keyOpt.toString(16)
+      const index = await this.blockIndex()
```
acolytec3, line 58: I can't remember why we did things this way before but why do we have to get a local reference to the block index?  And why is it on the base network class and not just history?  At a minimum, let's replace this sort of getter function and just expose the blockIndex publicly, or else write some useful wrappers around it for getting the block hash/number from it's mate.

packages/portalnetwork/src/networks/history/history.ts
```diff
@@ -70,20 +84,19 @@ export class HistoryNetwork extends BaseNetwork {
...
-      ? asBytes
-        ? header
-        : BlockHeader.fromRLPSerializedHeader(header, { setHardfork: true })
-      : undefined
+  ): Promise<undefined | (Uint8Array | BlockHeader)> => {
+    const contentKey =
+      'blockHash' in opt
+        ? getContentKey(HistoryNetworkContentType.BlockHeader, opt.blockHash)
+        : getContentKey(HistoryNetworkContentType.BlockHeaderByNumber, opt.blockNumber)
+    const value = await this.findContentLocally(fromHexString(contentKey))
```
acolytec3, line 94: Note to self, really, really, figure out how we represent these things internally.  Too many conversions here.

packages/portalnetwork/src/networks/history/history.ts
```diff
@@ -131,7 +148,15 @@ export class HistoryNetwork extends BaseNetwork {
       }
       if (Array.isArray(proof.value)) {
         try {
-          verifyPreMergeHeaderProof(proof.value, contentHash, header.number)
+          if ('blockHash' in validation) {
+            verifyPreMergeHeaderProof(proof.value, validation.blockHash, header.number)
+          } else {
+            verifyPreMergeHeaderProof(
```
acolytec3, line 154: Here's a theoretical question for us down the road.  Why are we proving things against a reconstructed `epochAccumulator` if we don't even have access to them any more.  Not to be solved now but I'm just curious.

packages/portalnetwork/src/networks/networkDB.ts
```diff
@@ -149,13 +149,40 @@ export class NetworkDB {
...
       if (d > radius) {
-        toDelete.push([key, value])
-        await this.db.del(key)
+        // Before deleting BlockHeaderWithProof (0x00) -- Check if BlockHeaderByNumber contentKey is in radius
+        if (key.startsWith('0x00')) {
+          // First find the block number from block index
+          const blockHash = '0x' + key.slice(4)
+          const blockNumber = blockIndex.get(blockHash)!
+          const numberKey = Uint8Array.from([
+            0x03,
```
acolytec3, line 168: Can we replace this with the `HistoryNetworkContentType.BlockHeaderByNumber` enum value?

packages/portalnetwork/test/integration/eth/getBlockByNumber.spec.ts
```diff
@@ -30,87 +29,21 @@ const privateKeys = [
...
-      await node1.stop()
-      await node2.stop()
-    },
-    { timeout: 30000 },
-  )
-  it('should find a block using "latest" and "finalized"', async () => {
-    vi.useFakeTimers({ shouldAdvanceTime: true, shouldClearNativeTimers: true })
-    vi.setSystemTime(1693431998000)
+describe(
+  'retrieve block using number',
```
acolytec3, line 34: Did you intent to remove the `it('...)` here?

packages/portalnetwork/test/integration/eth/getBlockByNumber.spec.ts
```diff
@@ -120,101 +53,162 @@ describe('eth_getBlockByNumber', () => {
...
+      assert.equal(Number(retrieved!.header.number), 1000, 'retrieved expected header')
+    })
 
-    const beacon = node1.networks.get(NetworkId.BeaconChainNetwork) as BeaconLightClientNetwork
+    await node1.stop()
+    await node2.stop()
+  },
+  { timeout: 30000 },
+)
+describe('should find a block using "latest" and "finalized"', async () => {
```
acolytec3, line 95: We need there to be `it('should...)` inside of these describe statements or we're going to have async weirdness.  `describe` should never be async as far as I remember from previous struggles with async stuff in `vitest`.

packages/portalnetwork/test/integration/integration.spec.ts
```diff
@@ -49,262 +49,226 @@ const testHashStrings: string[] = testHashes.map((testHash: Uint8Array) => {
 })
 
 describe('gossip test', async () => {
```
acolytec3, line 51: Let's keep the `it('...)`

acolytec3 (changes requested):
I think this generally looks great.  If we can clean up the tests so we don't lose the `describe('some test suite', () => { it('should do something', async ()...` layering, it should be good for now.  We can do a future PR to cleanup the blockindex getter stuff.

packages/portalnetwork/src/client/eth.ts
holgerd77: Yes

packages/portalnetwork/src/networks/history/history.ts
ScottyPoi: Not sure exactly, but I think because we imagined the blockIndex being updated / accessed by multiple networks
Regardless, the way we go about it can be nicer, for sure

packages/portalnetwork/src/networks/history/history.ts
ScottyPoi: Ah, this is overkill, actually.  The same thing is repeated in `findContentLocally`

packages/portalnetwork/src/networks/history/util.ts
ScottyPoi: yes

packages/portalnetwork/src/networks/history/util.ts
ScottyPoi: "the main driver is the db keys"

yes, I think that is true.  i know that we stuck with strings as db keys for a reason, but that is something we should revisit for this exact reason

packages/portalnetwork/test/integration/eth/getBlockByNumber.spec.ts
ScottyPoi: I think is is the other way around (async `it()` inside of non-async `describe`) that causes issues.

This does have an `it` test, at the bottom, that checks the results after all the async stuff happens inside of `describe`

ScottyPoi:
> I think this generally looks great. If we can clean up the tests so we don't lose the `describe('some test suite', () => { it('should do something', async ()...` layering, it should be good for now. We can do a future PR to cleanup the blockindex getter stuff.

I think it works the other way around, no?  Like:
```typescript
describe('some test suite', async () => {
    const result = await somethingAsync()
    it('checks this result', () => {
        assert.equal(result, expected)    
    })
})
```

packages/portalnetwork/src/networks/history/history.ts
ScottyPoi: For now, I implemented the wrapper `this.blockNumberToHash`

packages/portalnetwork/test/integration/eth/getBlockByNumber.spec.ts
acolytec3: Oh, really?  Then maybe I've got things backwards.  Will take a look at vitest here.

acolytec3 (approved):
Good step forward.

acolytec3 (approved):

packages/portalnetwork/src/client/eth.ts
holgerd77: Ah, you are right, forgot about the custom crypto thing, forget about it!

PR #628, opened 2024-09-19
Reduce bytes / hex conversions of content keys

ScottyPoi:
### Issue:
Excessive need to use `hexToBytes` and `bytesToHex` conversion methods on `contentKey`, `contentId`, `blockHash`, and more.

### Cause:
Inconsistent design of functions.  Some functions expected keys to be strings, others expected Uint8Arrays.

### Solution:
Internally, `contentKeys` should always be `Uint8Array`, with the only conversion to and from strings happening in the DB manager.

`contentId` should always be an **unprefixed hex string**, matching the type of `enr.nodeId`, and the expected input of the imported **distance** function `distance(nodeId, contentId)`

### DIFF
--- a/packages/cli/src/rpc/modules/beacon.ts
+++ b/packages/cli/src/rpc/modules/beacon.ts
@@ -100,11 +100,9 @@ export class beacon {
     }
     const lookup = new ContentLookup(
       this._beacon,
-      fromHexString(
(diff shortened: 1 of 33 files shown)


acolytec3 (approved):
LGTM - I salute your attention to details

PR #630, opened 2024-09-20
portalnetwork: return closer ENRs

ScottyPoi:
Adds a filter to the ENR response to FINDCONTENT

Response should only include ENR's closer than our own.


acolytec3 (commented):
LGTM.  Maybe add a test to confirm it's working right and then go ahead and merge!

ScottyPoi:
This is tested in `history-mesh` hive tests.  
These tests pass before this PR, and currently fail on this branch.  

**do-not-merge until resolved**

ScottyPoi:
Closing without merging.  I do not believe this is a correct change.

PR #645, opened 2024-10-08
Export CLI config function from PortalNetwork

ScottyPoi:
This PR creates a new function exported from `portalnetwork` package which creates a configuration object for `PortalNetwork.create()` from an input of CLI options.

The code inside the function was taken from `cli/src/index.ts`, and now replaces the configuration setup in that file.

The purpose here is to export the logic from the `portalnetwork` package, so that it can be reused during EthJS client integration, and avoid duplicating code.


holgerd77:
Congrats! 👍🙂

PR #649, opened 2024-10-14
StateNetwork refactor

ScottyPoi:
StateNetwork refactor and restructuring

Structure:
 - `StateNetwork` class
   -  `store`, `findContentLocally`, `sendFindContent`
   - helper methods for `store` 
     - `receive[...]Offer`, `storeInterested[...]Nodes`, `forward[...]Offer`
   - `this.manager: StateManager`
 - `StateManager` class
   - State lookup methods matching RPC methods
     - `getAccount` 
     - `getStorageAt` 
     - `getCode`
     - `getBalance`, `getNonce`, etc.
   - `this.trie: PortalTrie`
 - `PortalTrie` class
   -  wrapper methods around trie walking and node lookup
     - `find[...]Path(...)`
     - `lookup[...]TrieNode(...)`
   - Fetches nodes with Network Lookup when not found locally 
   - `this.db: PortalTrieDB`
     - `Trie` database extension that can find nodes in local DB or temporary node cache
 - `PortalTrieDB` class
   - extends DB interface for **@ethereumjs/trie** 
   - `this.db` hooks directly to local `StateNetwork` DB
   - `this.temp` stores path nodes during trie walking 
   - `this.local` maps `nodeHash` to DB content key for locally stored nodes
 

--

### Distinction between new `StateManager` class and preexisting `UltralightStateManager` class

`StateManager` class is an *attribute* of `StateNetwork`.  It enables state content lookup via `StateNetwork` retrieval methods.  It exposes methods like `getAccount`, `getCode`, and `getStorageAt`.  

`UltralightStateManager` class implements `EVMStateManagerInterface` from `@ethereumjs/statemanager`.   It also contains state lookup methods like `getAccount` and `getContractCode`, and hooks into the methods provided in `state.manager` for state lookup.  
It also implements state update methods.  It can be used to instantiate an `EVM` during methods like  `eth_call`


packages/portalnetwork/src/networks/state/manager.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import { ContentLookup } from '../contentLookup.js'
+
+import { PortalTrieDB } from './portalTrie.js'
+import { StateNetworkContentKey } from './util.js'
+
+import type { StateNetwork } from './state.js'
+import type { Path } from '@ethereumjs/trie'
+
+export class StateManager {
```
acolytec3, line 12: Is this going to implement the EthJS Statemanager interface ultimately?

ScottyPoi: No.  This will remain a readonly state interface, meant to retrieve state via local and network content.  This is the non-bridge client version of a state network client.  a non-bridge node has no need for creating new state, it just needs the methods to walk tries using the network.

PR #659, opened 2024-10-24
Skip gossip of individual LC updates

acolytec3:
Ultralight internally uses a content type of `0x15` for individual `LightClientUpdate`s so we can construct any requested range.  These individual updates were accidentally getting gossiped to other nodes (which is against the beacon spec) so this skips that data type when gossiping new content and adds a test to confirm.


ScottyPoi (approved):
LGTM!

PR #661, opened 2024-10-25
Switch all bytes hex converters

acolytec3:
Migrates our bytes <-> hex conversion utilities to `@ethereumjs/util` versions


ScottyPoi:
:clap:

PR #662, opened 2024-10-30
Improve Node Lookup

ScottyPoi:
This PR refactors and improves the `NodeLookup` class, as well as the `network.bucketRefresh` method.

We remove the `setInterval` code from the `PortalNetwork` class that sets a 30 seconds interval for `bucketRefresh`.  Instead, networks will manage their own `refreshInterval`, and reset this interval whenever `bucketRefresh` is called.  The interval is dynamically set based on current network conditions.  An empty routing table will refresh every 10 seconds.  Once the table is 30% full, we increase the interval to 30 seconds.  When the table is > 80% full, we increase the interval to 60 seconds.  This results in faster routing table population on startup.

Instead of `setInterval`, `PortalNetwork` will call `network.startRefresh()` on startup and `network.stopRefresh()` on close.

`NodeLookup` class was significanly refactored.  Many of the changes were organizational, breaking down the longer methods into shorter, more focused functions.  Other changes include removing the calls to `sendPing` on new enrs, since this already happens in `sendFindNodes`, and was therefor extraneous.   

`NodeLookup` also monitors the actual bucket size as the process continues, and aborts early when the bucket is full.

A `LOOKUP_TIMEOUT` was implemented for each individual  query.  This eliminated an issue with promises hanging open when peers failed to respond, and ensures the process continues quickly.

`network.sendFindNodes` was refactored as well.  `sendPing` already performs `evictNode` on unresponsive peers, so it was unnecessary to also call `evictNode` in `sendFindNodes`.  The conditional to only PING unkonwn nodes was also removed, as PINGing all the nodes received in a NODES message also acts as another `livenessCheck` and helps purge disconnected peers.

### DIFF
--- a/packages/portalnetwork/src/networks/network.ts
+++ b/packages/portalnetwork/src/networks/network.ts
@@ -781,11 +781,11 @@ export abstract class BaseNetwork extends EventEmitter {
 
   /**
    * Follows below algorithm to refresh a bucket in the routing table
-   * 1: Look at your routing table and select all buckets at distance greater than 239 that are not full.
-   * 2: Select a number of buckets to refresh using this logic (48+ nodes known, refresh 1 bucket, 24+ nodes known,
-   * refresh half of not full buckets, <25 nodes known, refresh all not empty buckets
-   * 3: Randomly generate a NodeID that falls within each bucket to be refreshed.
-   * Do the random lookup on this node-id.
+   * 1. Select 4 closest non-full buckets to refresh
+   * 2. Select a random node at the distance of each bucket
+   * 3. Perform a NodeLookup for the random node
+   * 4. NodeLookup will recursively query peers for new nodes at the distance of the bucket
+   * 5. New nodes will be added to the routing table
    */
   public bucketRefresh = async () => {
     const now = Date.now()
--- a/packages/portalnetwork/src/networks/network.ts
+++ b/packages/portalnetwork/src/networks/network.ts
@@ -343,7 +343,7 @@ export abstract class BaseNetwork extends EventEmitter {
           // Ping node if not currently ignored by subnetwork routing table
           await Promise.allSettled(
             notIgnored.map((e) => {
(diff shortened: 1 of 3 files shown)


packages/portalnetwork/src/networks/network.ts
```diff
@@ -64,6 +64,10 @@ export abstract class BaseNetwork extends EventEmitter {
   public bridge: boolean
 
   portal: PortalNetwork
+  private lastRefreshTime: number = 0
+  private nextRefreshTimeout: ReturnType<typeof setTimeout> | null = null
+  private refreshInterval: number = 30000 // Start with 30s
```
acolytec3, line 69: Is this used anywhere?

packages/portalnetwork/src/networks/network.ts
```diff
@@ -797,31 +788,37 @@ export abstract class BaseNetwork extends EventEmitter {
    * Do the random lookup on this node-id.
```
acolytec3, line 788: Will you update these comments using the description from the PR to explain how the new refresh works?

acolytec3 (commented):
Looks great.  Couple of small comments but great rework.

packages/portalnetwork/src/networks/network.ts
ScottyPoi: Yes.  look into `bucketRefresh` and the new "interval" methods at the bottom of this file

packages/portalnetwork/src/networks/network.ts
ScottyPoi: :heavy_check_mark:

packages/portalnetwork/src/networks/network.ts
acolytec3: I think you can remove this line since you have the refresh interval options coded in the `getRefreshInterval` method at the bottom.

acolytec3 (approved):

PR #665, opened 2024-10-31
ContentLookup followup: fix uTP dstId value

ScottyPoi:
Fixes `uTP` response to `sendFindContent`

`sendFindContent` can work with either a `nodeId` or `ENR` string.

However, `uTP.handleNewRequest` requires a `nodeId`, and not an `ENR` string.

This updates the call to `.handleNewRequest` by ensuring that `enr.nodeId` is passed, instead of passing the same string passed to `sendFindContent`


acolytec3 (approved):
LGTM

PR #669, opened 2024-11-05
RPC: fix recursiveNodeLookup

ScottyPoi:
Bug unconvered through use of `glados-cartographer`

`portal_*RecursiveFindContent` was not returning the expected array of ENRs. 

Changes: 
1. NodeLookup now takes an optional `refresh` parameter
  - if true, `NodeLookup` will exit early when the target bucket is full
  - set to true in `bucketRefresh`
2. `NodeLookup` now tracks all found ENRs in a `Heap` called `this.foundNodes`
  - `Heap` is automatically sorted by distance to target 
3. `NodeLookup.startLookup()` now returns the closest 16 ENRs from `foundNodes` heap
  - As `ENR` string array (sorted)
4. RPC methods `portal_*RecursiveFindNodes` now returns ENR list from `NodeLookup.startLookup()`

This PR also implements the same method for `state` and `beacon` networks, which were previously without this method.

### DIFF
--- a/packages/cli/src/rpc/modules/portal.ts
+++ b/packages/cli/src/rpc/modules/portal.ts
@@ -195,6 +195,12 @@ export class portal {
     this.historyRecursiveFindNodes = middleware(this.historyRecursiveFindNodes.bind(this), 1, [
       [validators.dstId],
     ])
+    this.stateRecursiveFindNodes = middleware(this.stateRecursiveFindNodes.bind(this), 1, [
+      [validators.dstId],
+    ])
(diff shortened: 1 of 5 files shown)


ScottyPoi:
I didn't realize that this method was used by Glados.  We otherwise only use the `NodeLookup` class during `bucketRefresh`, and recent updates to `NodeLookup` were optimized for this purpose.  These optimizations may not be ideal for the `glados-cartographer` use of `NodeLookup`, which may be interested in tallying ENRs that exist in the network outside of the local routing table.

acolytec3 (approved):
I also added some fixes for the `*traceGetContent` work in the `contentLookup` folder related to `glados-audit`

Issue #672, opened 2024-11-13
Random error linked to KademliaRoutingTable.getValue 


foufrix:
Error : 

```shell
  ddb83:Portal:StateNetwork:NODES Received 7 ENRs from f:c0a75...5e3cd +0ms
  ddb83:Portal:StateNetwork content added for: 0x20240000009a924888773203bdeea3297b7456fd80212aac9a6cca64dc898f343290cb29f300 +575ms
  ddb83:Portal:uTP:readSocket:25374:CLOSE Closing connection to dedfdee9cfe1d391f17ad84652cca1e4cdb2d0cb113189dab98564d44bfc31e1 +0ms
  ddb83:Portal:uTP:readSocket:25374:CLOSE compile=false +0ms
file:///Users/raf/Documents/ethereum/ultralight/node_modules/@chainsafe/discv5/lib/kademlia/kademlia.js:156
        return bucket.getValue(id);
                      ^

TypeError: Cannot read properties of undefined (reading 'getValue')
    at KademliaRoutingTable.getValue (/Users/raf/Documents/ethereum/ultralight/node_modules/@chainsafe/discv5/src/kademlia/kademlia.ts:186:19)
    at Discv5.findEnr (/Users/raf/Documents/ethereum/ultralight/node_modules/@chainsafe/discv5/src/service/service.ts:365:31)
    at SessionService.handleWhoAreYouRequest (/Users/raf/Documents/ethereum/ultralight/node_modules/@chainsafe/discv5/src/service/service.ts:708:22)
    at SessionService.emit (node:events:519:28)
    at SessionService.emit (node:domain:488:12)
    at SessionService.handleMessage (/Users/raf/Documents/ethereum/ultralight/node_modules/@chainsafe/discv5/src/session/service.ts:597:12)
    at UDPTransportService.processInboundPacket (/Users/raf/Documents/ethereum/ultralight/node_modules/@chainsafe/discv5/src/session/service.ts:312:21)
    at UDPTransportService.emit (node:events:519:28)
    at UDPTransportService.emit (node:domain:488:12)
    at Socket.handleIncoming (/Users/raf/Documents/ethereum/ultralight/node_modules/@chainsafe/discv5/src/transport/udp.ts:153:10)
```

This error pops randomly, most of the time around 10 minutes after starting a node on devnet

[...]

PR #673, opened 2024-11-14
solve client crash of KademliaTable.getValue

foufrix:
solve #672 
Error is probably due to a race condition in `discv5`

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -87,15 +87,18 @@ const main = async () => {
   }
 
   process.on('uncaughtException', (err) => {
+    // discv5 is occasionally throwing an uncaught exception when handling
+    // incoming messages related to the handshake process.  These can safely be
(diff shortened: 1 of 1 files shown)


packages/cli/src/index.ts
```diff
@@ -86,13 +86,16 @@ const main = async () => {
...
 
-  process.on('SIGINT', async () => {
-    console.log('Caught close signal, shutting down...')
-    await portal.stop()
-    if (metricsServer?.listening === true) {
-      metricsServer.close()
-    }
-    process.exit()
+  process.on('uncaughtException', (err) => {
+    console.error('Uncaught Exception:', err)
```
acolytec3, line 90: This should be it's own listener and not wrap the `sigint` listener.  That's what is causing that memory leak warning you reported.

acolytec3 (approved):

PR #680, opened 2024-12-05
Switch UTP to use full ENRs

acolytec3:
Update our UTP implementation to use full ENRs instead of node Ids as the reference inside our uTP sockets.  Also, save all ENRs discovered on the network to an in memory `enrCache` and use that instead of looking up nodes in the routing table when sending and responding to messages.


acolytec3:
Superseded by 681.

PR #681, opened 2024-12-05
No testdouble

ScottyPoi:
Removes `testdouble` dependency and deletes `testdouble` based unit tests.


acolytec3:
LGTM.  Stupid napi bugs.

PR #685, opened 2024-12-06
Speed up bootstrap

acolytec3:
Fixes a small UX painpoint by having the bootstrap process ping all bootnodes simultaneously instead of one at a time on client startup.


ScottyPoi (approved):
Good catch

PR #688, opened 2024-12-09
uTP Packet Queue

ScottyPoi:
### Issue: 
- Many `uTP` streams may be opened at once with a given peer
- This means as many different `uTP` **sockets** trying to send packets to the same peer at the same time.
- `discv5` talkReq buffer ends up too full -- most `talkRequests` timeout while in the buffer

### Solution:
- #### RequestManager
  -  `PortalNetworkUtp` maintains map of `PeerId > RequestManager`
    - client.utp.requestManagers[peerId]
  - Manages all open requests **per peer**
    - `client.utp.requestManagers[peerId].requestMap: Record<RequestId, ContentRequest>`
  - Handles all packets sent to **PEER**
    - `client.utp.requestManagers[peerId].handlePacket`
  - Maintains `PacketQueue` for each **peer**
    - `client.utp.requestManagers[peerId].packetQueue: Array<Packet>`
- ### PacketQueue
  - Attribute of `RequestManager`  
  - Array of all packets sent to any open requests with a given peer
  - In order of arrival
    - SYN and RESET packets jump to front of the line
  - Processes packets 1 at a time, in order, through `ContentRequest` handlers
    - `this.requestMap[requestId].handleUtpPacket(packet)`

----

### Avoid sending other requests peers with active uTP transfers

- Code added to check if peers have active uTP transfers before including them in other operations.
  - NodeLookup
  - ContentLookup
  - BucketRefresh

### DIFF
--- a/packages/portalnetwork/src/networks/contentLookup.ts
+++ b/packages/portalnetwork/src/networks/contentLookup.ts
@@ -73,6 +73,10 @@ export class ContentLookup {
     // Sort known peers by distance to the content
     const nearest = this.network.routingTable.values()
     for (const enr of nearest) {
+      // // Skip if the node has an active uTP request
+      if (this.network.portal.uTP.hasRequests(enr.nodeId) === true) {
+        continue
+      }
       const dist = distance(enr.nodeId, this.contentId)
       this.lookupPeers.push({ enr, distance: Number(dist) })
       this.meta.set(enr.nodeId, { enr: enr.encodeTxt(), distance: bigIntToHex(dist) })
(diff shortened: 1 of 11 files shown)


packages/portalnetwork/src/networks/beacon/beacon.ts
```diff
@@ -51,6 +51,7 @@ import type { Debugger } from 'debug'
 import type { AcceptMessage, FindContentMessage, OfferMessage } from '../../wire/types.js'
 import type { ContentLookupResponse } from '../types.js'
 import type { BeaconChainNetworkConfig, HistoricalSummaries, LightClientForkName } from './types.js'
+import type { INodeAddress } from '@chainsafe/discv5/lib/session/nodeInfo.js';
```
acolytec3, line 54: I'm really surprised this works since discv5 has an exports map.  Could we maybe just replicate this in our own types so we don't have to deal with a deep import like this?

acolytec3 (approved):
LGTM

PR #689, opened 2024-12-11
Move all offers to `gossipManager`

acolytec3:
This moves all `sendOffer` calls into the `gossipManager` class and makes it a property of the `BaseNetwork` class.  This should help to ensure we only have one outbound uTP stream open to another peer at a time.


ScottyPoi (approved):
LGTM.  Approved to merge if we have run this through the hive tests as well

Issue #691, opened 2024-12-15
Deal with data packet numbers at maxInt


acolytec3:
A recent test failure on hive indicates that we're not handling data packet incrementing correctly since the packet number is a 16 bit unsigned integer (so a maximum value of 65535) as visualized in the uTP [header diagram](https://www.bittorrent.org/beps/bep_0029.html).  When we increment [here](https://github.com/ethereumjs/ultralight/blob/cd85f4f52092faeacaa9b5f76db0fe52705c8f44/packages/portalnetwork/src/wire/utp/Socket/ContentReader.ts#L67), we need to check for maxint and then loop around to 0.

PR #692, opened 2024-12-16
Fix overflow issues in `seqNr` and `ackNr`

acolytec3:
This addresses #691. 

## The Problem
uTP sequence numbers are 16 bit unsigned integers so have a range of 0-65535.  When a sequence number reaches the maximum value, we would expect the next sequence number to be 0.  Ultralight was not checking for overflow and blithely continuing on in the sequence since we are using native Javascript `numbers` in our code rather than true 16-bit intengers.

## Solution
- Revise `contentWriter` so`dataChunks` are stored in an array of tuples of the form `[seqNr, bytes]` rather than an object keyed by the `seqNr` 
- Check for overflow in all sequence and ack number checks and updates
- Convert some `Object.keys` code to simple Array accesses (which should provide some small performance boost)
- Removes `writeSocket.dataNrs` as unnecessary

### DIFF
--- a/packages/portalnetwork/src/wire/utp/Packets/PacketHeader.ts
+++ b/packages/portalnetwork/src/wire/utp/Packets/PacketHeader.ts
@@ -2,7 +2,7 @@ import { VERSION } from '../Utils/constants.js'
 
 import { SelectiveAckHeaderExtension } from './Extensions.js'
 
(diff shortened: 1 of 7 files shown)


ScottyPoi (approved):
LGTM

PR #693, opened 2024-12-17
Use INodeAddress type in message handling

ScottyPoi:
Fixes issue caused by #682 / #684 

Message handling must use `INodeAddress` type, since `onTalkRequest` events from `discv5` do not always contain an `ENR`.

All network message handlers were updated to use `INodeAddress`, which is all that is needed for `sendTalkResponse`.
 
----

This gets a little tricky when a new `uTP` request is created.  All `uTP` messages happen via `talkReq`, which requires an `ENR` to send.  However, there are times when the `uTP` request needs to be instantiated before the peer's `ENR` has become available.  So a new `uTP` request can be created using either `INodeAddress` or `ENR`.  In cases where we use the `INodeAddress`, we optimistically hope that by the time we need to send the first `uTP` packet, the peer's `ENR` has become available.

### DIFF
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -406,7 +406,7 @@ export class PortalNetwork extends (EventEmitter as { new (): PortalNetworkEvent
     try {
       await this.uTP.handleUtpPacket(packetBuffer, src.nodeId)
     } catch (err: any) {
-      this.logger(err.message)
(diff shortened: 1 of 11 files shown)


ScottyPoi:
note: the last commit `56dfc59`, is a temporary patch for code that is currently getting refactored in a separate open PR.

acolytec3 (approved):

PR #694, opened 2024-12-17
client: fix config object issue

ScottyPoi:
This fixes an issue where `discv5` was not being instantiated with configuration options set during `PortalNetwork.create`.

We define a `defaultConfig` object for `PortalNetwork.create`, and modify it using args passed to the function like this:

`const config = { ...defaultConfig, ...opts.config }`

However, this was not effectively combining configuration options set specifically for discv5 which were in nested object `config.config`.

To fix this, we add this line after the one above: 

`config.config = { ...defaultConfig.config, ...opts.config.config }`

This solves the issue, and now `discv5` starts with the expected configuration.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -28,6 +28,8 @@ const main = async () => {
 
   const portal = await PortalNetwork.create(portalConfig)
 
+  log(`discv5Config: ${JSON.stringify(portal.discv5['config'], null, 2)}`)
+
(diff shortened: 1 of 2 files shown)


acolytec3 (approved):
LGTM

acolytec3 (commented):
Added one additional `opt` update where the `shouldRefresh` opt wasn't being populated in the `create` function.

PR #695, opened 2024-12-18
network.addBootNode: Check for pong before follup requests

ScottyPoi:
`BaseNetwork` method `addBootNode` first Pings a boodnode, then sends FindNodes request for non-full buckets.

This method was missing a conditional which first checked if the PING was successful or not (PONG received).  This led to many unnecessary requests being sent towards peers we already knew to be unresponsive.

This adds the necessary conditional `(if pong !== undefined) {...}`, and also corrects the requests to seek nodes from non-full buckets, not just empty buckets.


acolytec3 (approved):
LGTM

PR #699, opened 2024-12-23
PortalNetwork: Implement UltralightProvider Enhancements for EIP-1193 JSON-RPC Support

cjustinobi:
#### Key Enhancements:
1. **EIP-1193 Method Support:**
   - Implements commonly used Ethereum JSON-RPC methods:
     - `eth_getBlockByHash`
     - `eth_getBlockByNumber`
     - `eth_getTransactionCount`
     - `eth_getCode`
     - `eth_getBalance`
     - `eth_getStorageAt`
     - `eth_call`

2. **Error Handling:**
   - Adds robust error codes and messages based on JSON-RPC specifications:
     - `4200` for unsupported methods.
     - `-32602` for invalid parameters.
     - `-32603` for internal errors.
     
 2. **Removed  unused methods from `provider.spec.ts`**

### DIFF
--- a/packages/cli/src/rpc/modules/discv5.ts
+++ b/packages/cli/src/rpc/modules/discv5.ts
@@ -127,7 +127,7 @@ export class discv5 {
     const [nodeId] = params
     const buckets = (this._client.discv5 as any).kbuckets.buckets
     for (let i = 255; i > -1; i--) {
-      (buckets[i] as any).removeById(nodeId)
+      ;(buckets[i] as any).removeById(nodeId)
     }
     const removed = this._client.discv5.findEnr(nodeId)
     return removed === undefined ? true : false
--- a/packages/cli/src/rpc/modules/discv5.ts
+++ b/packages/cli/src/rpc/modules/discv5.ts
@@ -127,7 +127,7 @@ export class discv5 {
     const [nodeId] = params
     const buckets = (this._client.discv5 as any).kbuckets.buckets
     for (let i = 255; i > -1; i--) {
-      ;(buckets[i] as any).removeById(nodeId)
+      (buckets[i] as any).removeById(nodeId)
     }
     const removed = this._client.discv5.findEnr(nodeId)
     return removed === undefined ? true : false
--- a/packages/cli/scripts/testProvider.ts
+++ b/packages/cli/scripts/testProvider.ts
@@ -1,19 +1,27 @@
 import { UltralightProvider } from '../../portalnetwork/src/client/provider'
+import { NetworkId } from '../../portalnetwork/src/networks/types'
 
 const testBlockHash = '0x95b0950557cbc3e6647766adb719f80f7c7d192f4429b6026cdbd2cbe6a64294'
 const testContract = '0x6b175474e89094c44da98b954eedeac495271d0f'
 const testStorage = '0x0000000000000000000000000000000000000000000000000000000000000000'
-const historicalBlock = 31591
+const historicalBlock = 1048576
 
-async function findHistoricalAccount(provider: UltralightProvider, blockNumber: number): Promise<string | null> {
+async function findHistoricalAccount(
+  provider: UltralightProvider,
+  blockNumber: number,
+): Promise<string | null> {
   try {
     const block: any = await provider.request({
       method: 'eth_getBlockByNumber',
-      params: [blockNumber, true]
+      params: [blockNumber, true],
     })
 
-    if (block && block.result && block.result.transactions && block.result.transactions.length > 0) {
-
+    if (
+      block &&
+      block.result &&
+      block.result.transactions &&
+      block.result.transactions.length > 0
+    ) {
       const contractCreation = block.result.transactions.find((tx: any) => !tx.to)
       if (contractCreation) {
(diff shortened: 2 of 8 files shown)


packages/cli/src/rpc/modules/discv5.ts
```diff
@@ -127,7 +127,7 @@ export class discv5 {
     const [nodeId] = params
     const buckets = (this._client.discv5 as any).kbuckets.buckets
     for (let i = 255; i > -1; i--) {
-      ;(buckets[i] as any).removeById(nodeId)
+      (buckets[i] as any).removeById(nodeId)
```
acolytec3, line 130: Please remove this.  This isn't part of this PR.

acolytec3 (commented):
This is looking great so far.  One thing we need to address is that the provider should return JSON-RPC formatted data and not just the raw block/transaction/state data our `portalnetwork.ETH` module returns.  If you want to do that work as well, let me know and I can provide guidance on next steps or else I can do that in a subsequent PR.

Basically, we need to use [these helpers](https://github.com/ethereumjs/ethereumjs-monorepo/blob/3f0f3b506a0d12f5044b220bd533089c11b38bb6/packages/client/src/rpc/modules/eth.ts#L111) with some slight  edits that I can outline.

cjustinobi:
> This is looking great so far. One thing we need to address is that the provider should return JSON-RPC formatted data and not just the raw block/transaction/state data our `portalnetwork.ETH` module returns. If you want to do that work as well, let me know and I can provide guidance on next steps or else I can do that in a subsequent PR.
> 
> Basically, we need to use [these helpers](https://github.com/ethereumjs/ethereumjs-monorepo/blob/3f0f3b506a0d12f5044b220bd533089c11b38bb6/packages/client/src/rpc/modules/eth.ts#L111) with some slight edits that I can outline.

Please go ahead with the guidance. Would love to handle as well

acolytec3:
To make the provider fully EIP compliant, each of the supported methods needs to return the values formatted to match the Ethereum JSON-RPC specification.  So for each method listed in the PR description, you need to verify that the values returned by that method correspond to the [JSON-RPC spec](https://ethereum.github.io/execution-apis/api-documentation/).  

For the methods that return blocks, you can copy the the `toJSONRPCBlock` from the `ethereumjs-monorepo` client helpers [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/e2d2cb7603a53a0fa518c70c81c8a14e8dabfbe0/packages/client/src/rpc/modules/eth.ts#L111).

You'll need to update the `toJSONRPCBlock` method to remove the `chainId` parameter and then remove the `totalDifficulty` calculation.  Instead, just use the `difficulty` from the block header object as that field in the return value. 

For the other methods (`getCode`, `getBalance`, `getStorage`, `call`), you can reference the corresponding methods in the `eth` module in the ethereumjs client RPC implementation as a reference point for how to format the return values (or just reference the official spec).    

Also, if you can update your test script to make calls to these additional methods, that would be great.  For the state related methods, you'll have to pick an account that was present in the Ethereum state in ancient chain history (so maybe before block 10000000) since the state network doesn't fully support current state yet.  Let me know if you need help with this part of it.

cjustinobi:
> To make the provider fully EIP compliant, each of the supported methods needs to return the values formatted to match the Ethereum JSON-RPC specification. So for each method listed in the PR description, you need to verify that the values returned by that method correspond to the [JSON-RPC spec](https://ethereum.github.io/execution-apis/api-documentation/).
> 
> For the methods that return blocks, you can copy the the `toJSONRPCBlock` from the `ethereumjs-monorepo` client helpers [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/e2d2cb7603a53a0fa518c70c81c8a14e8dabfbe0/packages/client/src/rpc/modules/eth.ts#L111).
> 
> You'll need to update the `toJSONRPCBlock` method to remove the `chainId` parameter and then remove the `totalDifficulty` calculation. Instead, just use the `difficulty` from the block header object as that field in the return value.
> 
> For the other methods (`getCode`, `getBalance`, `getStorage`, `call`), you can reference the corresponding methods in the `eth` module in the ethereumjs client RPC implementation as a reference point for how to format the return values (or just reference the official spec).
> 
> Also, if you can update your test script to make calls to these additional methods, that would be great. For the state related methods, you'll have to pick an account that was present in the Ethereum state in ancient chain history (so maybe before block 10000000) since the state network doesn't fully support current state yet. Let me know if you need help with this part of it.

Got it

acolytec3 (commented):
I've pushed a commit that fixes a configuration thing I forgot about with state network and then cleans up the parameter typing within `provider.ts`.

`getBlockByNumber` isn't reliably returning a block yet so I need do some additional testing to see why that is since we should generally always be able to retrieve block by number.  The state network calls still aren't working right so I need to also research that a bit further since state network stuff has reliably worked for me in the past.

acolytec3 (approved):
LGTM.  Thanks for your work on this one!

I found one small bug in one of our `portal.ETH` initializers I've fixed here as well as some response formatting on the provider methods.  This has been a huge help though so thank you!

PR #702, opened 2025-01-10
Utp packet heap

ScottyPoi:
Replaces uTP Packet Queue array with `Heap` from `heap-js`.

Heap will sort packets waiting in queue so that out-of-order packets are reordered.  This will cut back on unnecessary `Selective Ack` responses.

When receiving streams, especially from `Trin` nodes, we often receive pairs of packets with identical timestamps, and regularly end up processing them in the wrong order.  
  - For example, we receive packets 1 and 2 simultaneously, but process packet 2 then packet 1.  Next we recieve 3 and 4, but process 4 before 3.
  - This results in a lot of extra packets being exchanged due the the `Selective Ack` function.  We report every other packet at missing, even though it isn't.  So We get multiples of every other packet and send extra `ACK` packets.

The `processCurrentPacket` method was updated to skip repeat packets, and to requeue packets that arrive ahead of the current reader position.  This update allows the packet-pair phenomenon to self correct.

Also updated and improved some logs, and added a method to report the total number of open uTP streams.

### DIFF
--- a/packages/portalnetwork/src/wire/utp/PortalNetworkUtp/requestManager.ts
+++ b/packages/portalnetwork/src/wire/utp/PortalNetworkUtp/requestManager.ts
@@ -1,10 +1,9 @@
 import type { ContentRequest } from "./ContentRequest.js";
-import { Packet , PacketType, UtpSocketType } from "../Packets/index.js";
+import { Packet, PacketType, UtpSocketType } from "../Packets/index.js";
 import type { Debugger } from "debug";
-import type {Comparator} from "heap-js";
-import { Heap} from "heap-js";
-
-type RequestId = number
+import type { Comparator } from "heap-js";
+import { Heap } from "heap-js";
+import { MAX_IN_FLIGHT_PACKETS, type RequestId } from "./types.js";
 
 const packetComparator: Comparator<Packet<PacketType>> = (a: Packet<PacketType>, b: Packet<PacketType>) => {
     // If packets belong to the same connection, sort by sequence number
@@ -14,9 +13,6 @@ const packetComparator: Comparator<Packet<PacketType>> = (a: Packet<PacketType>,
     // Otherwise, sort by timestamp
     return a.header.timestampMicroseconds - b.header.timestampMicroseconds;
 }
-
-const MAX_IN_FLIGHT_PACKETS = 3
-
 export class RequestManager {
     peerId: string
     requestMap: Record<RequestId, ContentRequest>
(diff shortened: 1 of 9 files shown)


packages/portalnetwork/src/wire/utp/PortalNetworkUtp/requestManager.ts
```diff
@@ -50,43 +61,52 @@ export class RequestManager {
...
+            this.currentPacket = this.packetHeap.pop()
             await this.processCurrentPacket()
             return
         }
+        if (request.socket.type === UtpSocketType.READ && request.socket.reader !== undefined) {
+            if (this.currentPacket.header.seqNr < request.socket.reader!.nextDataNr) {
+                this.logger.extend('PROCESS_CURRENT_PACKET')(`Packet ${this.currentPacket.header.seqNr} already processed.`)
+                this.currentPacket = this.packetHeap.pop()
+                return this.processCurrentPacket()            
+            } else if (this.currentPacket.header.seqNr > request.socket.reader!.nextDataNr) {
```
acolytec3, line 101: This is what makes selective acks basically never happen, right?

ScottyPoi: Right -- this still lacks a way to respond to actual packet loss.  This will just happen infinitely (or till other side sends a RESET).

ScottyPoi: Commit 774053c adds some logic to process out of order packets if there are 3 or more packets for that request in the heap.  This preserves the reordering for the packet pair phenomenon, but still allows the `selective ack` response to happen if we get more than 3 packets ahead.

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/requestManager.ts
```diff
@@ -50,65 +86,79 @@ export class RequestManager {
...
+                if (this.getPacketCount(this.currentPacket.header.connectionId) < 3) {
```
acolytec3, line 127: Can we make this a constant with explanatory text so it's obvious what "3" is?

acolytec3 (changes requested):
Looks great, with one change for clarity needed I think. Otherwise, go ahead and merge and don't wait for an approval from me

acolytec3 (approved):
LGTM

PR #703, opened 2025-01-10
uTP:  concurrent request limit

ScottyPoi:
Limits the total number of active uTP streams to `MAX_CONCURRENT_UTP_STREAMS`.  

Limit defaults to 50.  Chosen to match default set by other clients.

New OFFERs will be rejected while number of active requests is at this limit.


acolytec3 (approved):
LGTM

PR #706, opened 2025-01-14
Ping/Pong Extensions

ScottyPoi:
Implements changes in https://github.com/ethereum/portal-network-specs/pull/348

Redefines the `CustomPayload` of PING and PONG messages to include a `type` parameter.

This allows different types of PING messages to provide / request different kinds of data from a peer.

Currently this includes:

* type 0: Client Info and Capabilities
* type 1: Basic Radius
* type 2: History Radius

Clients will respond with a PONG of the same type, or:
* type 65535: Error Response


packages/portalnetwork/src/client/client.ts
```diff
@@ -171,6 +176,12 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
   constructor(opts: PortalNetworkOpts) {
     // eslint-disable-next-line constructor-super
     super()
+    this.clientInfo = {
+      clientName: 'ultralight',
+      clientVersionAndShortCommit: `${packageJson.version}-${opts.shortCommit ?? ''}`,
+      operatingSystemAndCpuArchitecture: `${process.platform}-${process.arch}`,
```
acolytec3, line 182: ideally, this is not something we should pull from a NodeJS builtin as it makes browser client stuff that much harder.  Can we just make this a required parameter to pass in to the constructor?

packages/portalnetwork/src/networks/network.ts
```diff
@@ -288,19 +364,53 @@ export abstract class BaseNetwork extends EventEmitter {
...
-    await this.sendPong(src, id)
+    const { type, payload } = CustomPayloadExtensionsFormat.deserialize(pingMessage.customPayload)
+    let pongPayload: Uint8Array
+    switch (type) {
+      case PingPongPayloadExtensions.CLIENT_INFO_RADIUS_AND_CAPABILITIES: {
+        pongPayload = this.pingPongPayload(type)
+        break
+      }
+      case PingPongPayloadExtensions.BASIC_RADIUS_PAYLOAD: {
+        if (this.networkId !== NetworkId.HistoryNetwork) {
```
acolytec3, line 378: Shouldn't we support basic radius on all networks?

acolytec3 (changes requested):
Looks good so far!  Couple of relatively small things in comments.

Other feedback already noted on discord.

packages/portalnetwork/src/networks/network.ts
ScottyPoi: Ah, whoops.  that should be in the HISTORY_RADIUS case

packages/cli/src/index.ts
```diff
@@ -26,6 +26,9 @@ const main = async () => {
 
   const portalConfig = await cliConfig(args)
 
+  portalConfig.operatingSystemAndCpuArchitecture = args.arch
+  portalConfig.shortCommit = args.commit ?? (await simpleGit().revparse(['HEAD'])).substring(0, 7)
```
acolytec3, line 30: I have to ask because I'm "that guy" but is it necessary to add a new dep for this or can we just execute a shell command for `git revparse` and assume that whoever is running this will have git installed locally?

acolytec3 (changes requested):
Looks good to me overall though I question whether we need a new dependency to grab the commit hash.  I think we can safely assume that anybody running this client locally will either be running it from a cloned git repo or in docker and I'm pretty sure git is installed in our docker image so we should able to just do an `exec('git rev-parse HEAD')`  or something to that effect and read the output instead of requiring a new dependency for this.  I know I'm being persnickety but just my 2 gwei.  Happy to be persuaded otherwise.

packages/cli/src/index.ts
ScottyPoi: Changed it

ScottyPoi:
commit 58bd756 implements changes from https://github.com/ethereum/portal-network-specs/pull/358

unit tests will fail until test vectors are updated

ScottyPoi:
Passing tests with test vectors from: https://github.com/ethereum/portal-network-specs/pull/357/files

packages/portalnetwork/test/integration/pingpong.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+      payloadType: PingPongPayloadExtensions.HISTORY_RADIUS_PAYLOAD,
+      customPayload: pingPayload,
+    }
+    const network1NodeAddress: INodeAddress = {
+      nodeId: node1.discv5.enr.nodeId,
+      socketAddr: initMa,
+    }
+    const pong = await network2.handlePing(network1NodeAddress, 1234n, pingMsg)
+    const pongMsg = PortalWireMessageType.deserialize(pong)
+    // const pongPayload = HistoryRadius.deserialize(
```
acolytec3, line 84: Why is this commented out?  Shouldn't we validate the content of the error payload too?

acolytec3 (approved):
LGTM

Dockerfile
```diff
@@ -7,6 +7,7 @@ RUN ln -s /lib/libc.musl-x86_64.so.1 /lib/ld-linux-x86-64.so.2
 
 WORKDIR /ultralight
 
+COPY .git .git
```
KolbyML, line 10:
```suggestion
COPY .git /.git
```
This should copy the .git folder into the root directory, because ultralight is being ran from the / directory so node will look for `.git` in the / directory

KolbyML:
```
    const portal = new PortalNetwork({
      config,
      bootnodes,
      db: opts.db,
      supportedNetworks: opts.supportedNetworks ?? [
        { networkId: NetworkId.HistoryNetwork, maxStorage: 1024 },
      ],
      dbSize: dbSize as () => Promise<number>,
      metrics: opts.metrics,
      trustedBlockRoot: opts.trustedBlockRoot,
      eventLog: opts.eventLog,
      utpTimeout: opts.utpTimeout,
      gossipCount: opts.gossipCount,
      shouldRefresh: opts.shouldRefresh,
    })
```

needs to be updated to 

```
    const portal = new PortalNetwork({
      config,
      bootnodes,
      db: opts.db,
      supportedNetworks: opts.supportedNetworks ?? [
        { networkId: NetworkId.HistoryNetwork, maxStorage: 1024 },
      ],
      dbSize: dbSize as () => Promise<number>,
      metrics: opts.metrics,
      trustedBlockRoot: opts.trustedBlockRoot,
      eventLog: opts.eventLog,
      utpTimeout: opts.utpTimeout,
      gossipCount: opts.gossipCount,
      shouldRefresh: opts.shouldRefresh,
      operatingSystemAndCpuArchitecture: opts.operatingSystemAndCpuArchitecture,
      shortCommit: opts.shortCommit,
    })
```    

My tests pass with this change

PR #710, opened 2025-01-16
Post capella block verification

acolytec3:
- Adds parsing of blocks from `era` files and better error handling
- Adds a new post capella bridge script for constructing block header proofs rooted in `HistoricalSummaries`
- Allow verification of post-capella headers based on having the `HistoricalSummaries` present
- Cleans up header verification logic in `HistoryNetwork.validateHeader`
  - Verify that header data constructs a header with the same hash as the provided block hash
  - Switch header proof verification utilities to return boolean values instead of throwing


packages/era/src/helpers.ts
```diff
@@ -55,6 +56,26 @@ export const deserializeE2Store = (bytes: Uint8Array): e2StoreEntry[] => {
   return entries
 }
 
+
+function littleEndianToBigInt(bytes: Uint8Array) {
```
ScottyPoi, line 60: does the ethjs `bytesToBigInt` function not work here?  It has a `littleEndian: boolean = false` parameter

ScottyPoi (approved):
Approving with minimal feedback.  Some comments in the code explaining what is happening would be helpful.

packages/era/src/helpers.ts
acolytec3: When I switch to `bytesToBigInt`, my tests start failing so something's awry.  Will research a bit more.

ScottyPoi (approved):

PR #712, opened 2025-01-16
Blacklist via RateLimiter

ScottyPoi:
Implements a solution for blacklisting nodes at the UDP level.

The `UDPTransportService` class is the default `transport` layer used by discv5 to interact with UDP sockets.

This class can be constructed with an object called `RateLimiter`.  If `transport.rateLimiter` exists, `transport.handleIncomingPacket` will check `rateLimiter.allowEncodedPacket` before processing incoming packets.

This PR introduces a custom `RateLimiter` which will maintain a set of blacklisted IP addresses.  When `tranport.handleIncomingPacket` calls `rateLimiter.allowEncodedPacket`, the `RateLimiter` will check this set of blacklisted IP addresses before allowing the packet to be processed.

This now allows us to blacklist peers on the UDP level.

Previously, Ultralight could add a peer to `routingTable.ignored`, which prevents us from sending new requests to peers that have been unresponsive.  `RateLimiter` allows us to block **incoming** messages from banned peers.

Currently the timeout for blacklisting is set at 24 hours.

### DIFF
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -32,6 +32,7 @@ import type {
   PortalNetworkOpts,
 } from './types.js'
 import { MessageCodes, PortalWireMessageType } from '../wire/types.js'
+import { RateLimiter } from '../transports/rateLimiter.js'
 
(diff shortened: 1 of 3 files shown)


acolytec3 (approved):
LGTM.  :shipit:

PR #714, opened 2025-01-17
uTP:  Handle cases of unexpected packets

ScottyPoi:
This PR implements solutions for cases where Ultralight receives a `uTP` packet for a closed or non-existent connection.

Previously Ultralight would ignore these packets on the `uTP` level.  The packet arrives via `talkReq`, UL responds with `talkResp`, and sends the packet to the `uTP` packet queue to process.  During `handleCurrentPacket`, `uTP` would find no open request for the packet, and simply throw it away and move on to the next packet in the queue.  

There are generally 3 scenarios where this may occur:

1. We completed a `uTP` stream, sending a FIN-ACK packet to close the stream, but our peer failed to receive this packet.  Since the peer still believes the connection to be open, they resend the FIN packet one or more times, attempting to complete the transfer.  
2. We aborted a `uTP` stream, sending a RESET packet to close the stream, but our peer failed to receive the RESET.  Unaware that we have closed our end, they continue to send packets, awaiting a reply that will never come.
3. A malicious peer sends bogus `uTP` packets as a DoS attack.

In the first 2 scenarios, we are dealing with a well-meaning peer who is behaving as expected, unaware that packet-loss has corrupted the connection.  The proper response in these cases should be to send this peer a RESET packet, so that they know to close the connection on their side.

In the attack scenario, we waste resources by processing the malicious packets and responding with our own RESET packets.  For each malicious packet, we would be responding with 2 of our own packets (talkResp, and RESET).  With no safeguards, a bad actor could send infinite malicious packets as a DoS attack.

PR: #712  enabled Ultralight to **BLACKLIST** peers on the UDP level.  UDP packets from a blacklisted peer will never be processed by the client.

This PR modifies `uTP` to respond with BLACKLIST when an unexpected packet can't be traced to a closed connection, and to respond with RESET when a packet arrives for a previously closed stream.   An exception to this is when receiving a RESET packet for an already closed connection.  In that case we do not send a RESET, as this could result in an infinite RESET loop between peers.

This should improve network health by closing connections that linger open, and improve client health by safeguarding against this DoS attack vector.

### DIFF
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -391,11 +391,6 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
   ) => {
     this.metrics?.totalBytesReceived.inc(message.request.length)
     if (bytesToHex(message.protocol) === NetworkId.UTPNetwork) {
-      if (!this.discv5.findEnr(nodeAddress.nodeId)) {
-        this.logger.extend('TalkReq').extend('error')(`Received uTP packet from unknown node: ${nodeAddress.nodeId}. Adding to blacklist.`)
-        this.addToBlackList(nodeAddress.socketAddr)
-        return
-      }
       await this.handleUTP(nodeAddress, message, message.request)
       return
     }
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -391,6 +391,11 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
   ) => {
     this.metrics?.totalBytesReceived.inc(message.request.length)
     if (bytesToHex(message.protocol) === NetworkId.UTPNetwork) {
+      if (!this.discv5.findEnr(nodeAddress.nodeId)) {
+        this.logger.extend('TalkReq').extend('error')(`Received uTP packet from unknown node: ${nodeAddress.nodeId}. Adding to blacklist.`)
+        this.addToBlackList(nodeAddress.socketAddr)
+        return
+      }
       await this.handleUTP(nodeAddress, message, message.request)
       return
     }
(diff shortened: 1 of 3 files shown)


packages/portalnetwork/src/client/client.ts
```diff
@@ -391,6 +391,11 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
   ) => {
     this.metrics?.totalBytesReceived.inc(message.request.length)
     if (bytesToHex(message.protocol) === NetworkId.UTPNetwork) {
+      if (!this.discv5.findEnr(nodeAddress.nodeId)) {
```
acolytec3, line 394: This doesn't seem like it should work correctly.  Why are we using the discv5 routing table to lookup nodes that are very likely not in the discv5 routing table but would be in one of the network routing tables (assuming it's a real node that we are interacting with and likely have a uTP socket with)?

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/requestManager.ts
```diff
@@ -73,29 +74,55 @@ export class RequestManager {
...
+        )
+        throw new Error(`REQUEST_NOT_FOUND`)
+      }
+      default: {
+        if (
+          packet.header.pType === PacketType.ST_SYN ||
+          packet.header.pType === PacketType.ST_RESET
+        ) {
+          this.logger.extend('HANDLE_PACKET')(
+            `Processing SYN packet for Req:${packet.header.connectionId}`,
```
acolytec3, line 109:
```suggestion
            const packetType = packet.header.pType === PacketType.ST_SYN ? 'SYN' : 'RESET'
          this.logger.extend('HANDLE_PACKET')(
            `Processing ${packetType} packet for Req:${packet.header.connectionId}`,
```

This log message should reflect the actual packet type being reflected (since it could be either)

acolytec3 (changes requested):
Looks good as far as it goes.  Not sure I understand the blacklist at the `onTalkResp` level though so want to understand that before we merge.

packages/portalnetwork/src/client/client.ts
ScottyPoi: Ah, yeah you're right.  checking the routing table doesn't necessarily tell us whether we know this peer or not.

ScottyPoi: Deleted this.  It was overkill anyway, as the utp handler will check if for any history with that peer

acolytec3 (approved):

PR #715, opened 2025-01-17
Update vitest to v3.0.2

cjustinobi:
### Changes

 - Updated `vitest` and `@vitest/ui` dependency to ^3.0.2.
 - Added `@vitest/coverage-v8` for test coverage
 - Added `test:ui` script for web view of the coverage


cjustinobi:
The coverage was failing at the reference of `portal-spec-tests` submodule. Had to exclude
it for the time being.

packages/portalnetwork/vitest.config.unit.ts
```diff
@@ -3,7 +3,18 @@ import { defineConfig } from 'vitest/config'
 export default defineConfig({
   test: {
     silent: true,
-    exclude: ['test/integration'],
+    exclude: [
+      'test/integration',
+      'test/networks/history/portalSpecTests.spec.ts',
+      'test/networks/state/content.spec.ts',
```
acolytec3, line 9: Why these new exclusions?

jochem-brouwer:
Vitest just got an update to 3.0.x so it might be worth immediately updating to that :)

packages/portalnetwork/vitest.config.unit.ts
acolytec3: I think you just needed to run `git submodule -- init` or something.  They all are passing locally for me so I'm going to revert this change.

ScottyPoi:
> Vitest just got an update to 3.0.x so it might be worth immediately updating to that :)

I second this!

acolytec3:
I'm not 100% sure what's going on but now when I run the tests locally, some of them hang.  We might need to do some more research on why they're hanging before we upgrade.  I'll put this on my todo list for next week.  

@cjustinobi could you go ahead and update to vitest v3 like @jochem-brouwer and then we can just do this upgrade once, once we figure out why tests are hanging.

cjustinobi:
> I'm not 100% sure what's going on but now when I run the tests locally, some of them hang. We might need to do some more research on why they're hanging before we upgrade. I'll put this on my todo list for next week.
> 
> @cjustinobi could you go ahead and update to vitest v3 like @jochem-brouwer suggested and then we can just do this upgrade once, once we figure out why tests are hanging.

It seems the issue is coming from `packages/portalnetwork/test/integration/eth/getBlockByNumber.spec.ts`

acolytec3 (approved):
LGTM

acolytec3 (commented):
Thanks for this!

PR #717, opened 2025-01-22
Get block by hash loop

### LINKED ISSUE #716: `ETH.getBlockByHash` goes into infinite loop if block not found
In the course of working on #710, I discovered that our `client.ETH.getBlockByHash` implementation can go into an infinite loop if we manage to index a blockhash <> block number mapping but where the block isn't stored correctly in the DB.  At the moment, this appears to mainly be an issue when the node doesn't have any external peer connections.

The issue occurs in this [code branch](https://github.com/ethereumjs/ultralight/blob/18ba2b833acda8b82618fb36b5f64eeab6269063/packages/portalnetwork/src/client/eth.ts#L79).  We don't find the block by the hash, but we do have a mapping of the block hash to block number so we try looking it up by number.  If we don't find it by number, [but do have the number to hash mapping](https://github.com/ethereumjs/ultralight/blob/18ba2b833acda8b82618fb36b5f64eeab6269063/packages/portalnetwork/src/client/eth.ts#L180), we then try to look it up by hash again.

acolytec3:
Removes fallback pattern where `getBlockByHash` and `getBlockByNumber` would try to retrieve via number or hash respectively if the direct retrieval failed as this caused infinite looping searches for a block.

Fixes #716

### DIFF
--- a/packages/portalnetwork/src/client/eth.ts
+++ b/packages/portalnetwork/src/client/eth.ts
@@ -75,12 +75,6 @@ export class ETH {
       lookupResponse = await lookup.startLookup()
       this.logger.extend('getBlockByHash')(lookupResponse)
       if (!lookupResponse || !('content' in lookupResponse)) {
-        // Header not found by hash, try to find by number if known
-        const blockNumber = this.history!.blockHashToNumber(blockHash)
-        if (blockNumber !== undefined) {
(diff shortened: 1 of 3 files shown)


ScottyPoi (approved):
LGTM

PR #721, opened 2025-01-24
WIP: REPL Interface for Portal Network

cjustinobi:
This PR introduces a REPL interface for interacting with the Portal Network, enabling developers to test and manage the network more effectively.

### Features:
**REPL Commands**:
   - `.ping <network>`: Send a ping request to the specified network (`history` or `state`).
   - `.findnodes <network> <enr> <distances>`: Locate nodes within a network using an ENR and distances.
   - `.findcontent <network> <enr> <contentKey>`: Search for content in a network using an ENR and content key.
   - `.offer <network> <enr> <contentKey> <contentValue>`: Offer content to a node in the specified network.
   - `.status`: Display the current status of the Portal Network and its components.

### DIFF
--- a/packages/portalnetwork/examples/portalClient.ts
+++ b/packages/portalnetwork/examples/portalClient.ts
@@ -1,0 +1,120 @@
+import { SignableENR, ENR } from '@chainsafe/enr'
+import { keys } from '@libp2p/crypto'
+import { multiaddr } from '@multiformats/multiaddr'
+import { PortalNetwork, NetworkId, TransportLayer, HistoryNetwork, StateNetwork } from '../../portalnetwork/src'
+import repl from 'repl'
+
+class PortalNetworkRepl {
+  private node?: PortalNetwork
+  private historyNetwork?: any
+  private stateNetwork?: any
+
+  async init(port = 9090): Promise<void> {
+    const privateKey = await keys.generateKeyPair('secp256k1')
+    const enr = SignableENR.createFromPrivateKey(privateKey)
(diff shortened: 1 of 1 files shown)


cjustinobi:
@acolytec3 I just want to make sure I'm in the right track. Please your feedback

acolytec3 (commented):
Looks good so far.  I pushed a couple of tweaks so it will actually connect to the network now.  You can also tune the logs it shows because the stuff you have now is a lot of logs.

cjustinobi:
Alright. Also, the methods I implemented so far are as a result of what I saw from [here](https://ethportal.net/concepts/protocols/portal_wire#relationship-to-json-rpc) Are there more stuff I could have done there?

acolytec3:
The full rpc spec is [here](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false)

I don't think we need to implement all of them but maybe like `addENR` would be helpful.

acolytec3 (approved):
LGTM.  We need a different (more accurate) mechanism for enforcing "first ping is type 0", maybe through an enr cache that we've played around with in the past.

Deejae69 (approved):

PR #724, opened 2025-01-29
Update Discv5 and Remove Buffer

ScottyPoi:
Updates `@chainsafe/discv5` to v.11.0.0
Updates `@chainsafe/enr` to v.5.0.0

Removes all remaining `Buffer` use.
Rewrites `Buffer` based code to use `Uint8Array`


acolytec3 (approved):
I guess this is okay.

PR #726, opened 2025-02-03
ENR Cache

ScottyPoi:
## ENR Cache + Peer tracking

Introduces a cache of peers as an attribute of `PortalNetwork` class.  Cache will track `ENR`'s of all contacted nodes, as well as `capabilities`, `client_info`, and other data for peer tracking / peer scoring.

This cache is independent of network and discv5 `RoutingTables`.  Peers tracked by the cache are not necessarily "in the routing table" for any network.

ENR info is kept updated during message handling.

Capabilities and other Ping/Pong payload info is updated during Ping/Pong.

The cache is stored in the DB on close, and rebuilt on startup if available.

### DIFF
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -544,24 +544,31 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
 
   public updateENRCache = (enrs: ENR[]) => {
     for (const enr of enrs) {
-      this.enrCache.set(enr.nodeId, enr)
+      this.enrCache.updateENR(enr)
     }
   }
 
   public findENR = (nodeId: string): ENR | undefined => {
-    return this.enrCache.get(nodeId) ?? this.discv5.findEnr(nodeId)
+    return this.enrCache.getENR(nodeId) ?? this.discv5.findEnr(nodeId)
   }
 
   public storeENRCache = async () => {
-    const cache = Array.from(this.enrCache.entries()).map(([nodeId, enr]) => [nodeId, enr.encodeTxt()])
+    const cache = Array.from(this.enrCache['peers'].values())
+      .filter((peer) => peer.enr !== undefined)
+      .map((peer) => {
+        return peer.enr!.encodeTxt()
+      })
     await this.db.put('enr_cache', JSON.stringify(cache))
   }
 
   public loadENRCache = async () => {
-    try { 
+    try {
       const storedEnrCache = await this.db.get('enr_cache')
       if (storedEnrCache) {
-        this.enrCache = new Map(JSON.parse(storedEnrCache))
+        const enrs = JSON.parse(storedEnrCache)
+        for (const enr of enrs) {
+          this.enrCache.updateENR(ENR.decodeTxt(enr))
+        }
       }
     } catch {
       // No action
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -431,9 +431,6 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
 
   private onTalkResp = (_: INodeAddress, src: ENR | null, message: ITalkRespMessage) => {
     this.metrics?.totalBytesReceived.inc(message.response.length)
-    if (src) {
-      this.updateENRCache([src])
-    }
   }
 
   /**
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -298,9 +298,13 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
     await this.discv5.start()
     await this.db.open()
     const storedIndex = await this.db.getBlockIndex()
-    const storedEnrCache = await this.db.get('enr_cache')
-    if (storedEnrCache) {
-      this.enrCache = new Map(JSON.parse(storedEnrCache))
+    try {
+      const storedEnrCache = await this.db.get('enr_cache')
+      if (storedEnrCache) {
+        this.enrCache = new Map(JSON.parse(storedEnrCache))
+      }
+    } catch {
+      // No action
     }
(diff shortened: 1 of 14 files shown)


packages/portalnetwork/src/client/client.ts
```diff
@@ -49,7 +49,7 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
   metrics: PortalNetworkMetrics | undefined
   logger: Debugger
   ETH: ETH
-
+  enrCache: Map<string, ENR>
```
acolytec3, line 52:
```suggestion
  enrCache: Map<NodeId, ENR>
```

I know this is purely semantics but it's a nice one to explicitly define how we're mapping to the ENR.

You'll also need to add the import from `@chainsafe/enr`

packages/portalnetwork/src/networks/contentLookup.ts
```diff
@@ -193,6 +193,7 @@ export class ContentLookup {
         this.logger(`received ${res.enrs.length} ENRs for closer nodes`)
         for (const enr of res.enrs) {
           const decodedEnr = ENR.decode(enr as Uint8Array)
+          this.network.portal.updateENRCache([decodedEnr])
```
acolytec3, line 196: Should we actually update it here or should we wait until we get a PONG from the node to know its a live node?

packages/portalnetwork/src/networks/nodeLookup.ts
```diff
@@ -82,6 +82,7 @@ export class NodeLookup {
 
         for (const enr of response.enrs) {
           const decodedEnr = ENR.decode(enr)
+          this.network.portal.updateENRCache([decodedEnr])
```
acolytec3, line 85: Same as content lookup.  Doesn't it make more sense to only add to the cache once we actually make contact with a node?

packages/portalnetwork/test/integration/pingpong.spec.ts
```diff
@@ -57,17 +57,17 @@ describe('PING/PONG', async () => {
...
-      await network1.sendPing(network2?.enr!.toENR(), 1)
-      assert.fail('should have failed')
-    } catch (e) {
-      assert.equal(
-        e.message,
-        'First PING message must be type 0: CLIENT_INFO_RADIUS_AND_CAPABILITIES.',
-      )
-    }
-  })
+  // it('should fail if first ping is not CLIENT_INFO_RADIUS_AND_CAPABILITIES', async () => {
```
acolytec3, line 60: Let's just go ahead and remove this if everyone agrees we shouldn't be dropping a connection because somebody else sent us a ping (that we think might be a first ping) that isn't type 0.

acolytec3 (commented):
Some small feedback but bigger general question.  Shouldn't we wait to add to the ENR cache once we have established a session with another node?  Holding onto ENRs for nodes we haven't established a session with yet doesn't seem super fruitful to me.

acolytec3 (commented):
I'll leave this as a general comment since it applies in a number of places but should we limit the times we update the `enrCache` to when it actually makes sense?  To my way of thinking, we should only update it in the scenarios where it makes sense to and not on every single TALKREQ/TALKRESP which is what it looks like we're doing now. Even if each individual update isn't that resource intensive, this is adding a ton of duplicative and potentially performance reducing overhead if we do it every time we send or receive a Portal Network message (which is what is going to be happening if we do it on every `sendPortalNetworkMessage` and every `onTalkReq`/`onTalkResp`.  
As such, doesn't it make sense to limit updating the cache to:
- Discovering a new peer (so in discv5 discovery and session established messages)
- When we receive a PING or send a PONG
- When we hit some error condition where we need to ban/ignore/score a peer

packages/portalnetwork/src/client/client.ts
```diff
@@ -421,8 +435,11 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
     await network.handle(message, nodeAddress)
   }
 
-  private onTalkResp = (_: any, __: any, message: ITalkRespMessage) => {
+  private onTalkResp = (_: INodeAddress, src: ENR | null, message: ITalkRespMessage) => {
     this.metrics?.totalBytesReceived.inc(message.response.length)
+    if (src) {
+      this.updateENRCache([src])
+    }
```
acolytec3, line 442: As above, remove
```suggestion
```

packages/portalnetwork/src/client/client.ts
```diff
@@ -464,11 +481,14 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
     networkId: NetworkId,
     utpMessage?: boolean,
   ): Promise<Uint8Array> => {
+    if (enr instanceof ENR) {
+      this.updateENRCache([enr])
+    }
```
acolytec3, line 486: As above, remove
```suggestion
```

packages/portalnetwork/src/networks/network.ts
```diff
@@ -328,16 +329,27 @@ export abstract class BaseNetwork extends EventEmitter {
...
           }
           case PingPongPayloadExtensions.HISTORY_RADIUS_PAYLOAD: {
             const { dataRadius } = HistoryRadius.deserialize(pongMessage.customPayload)
             this.routingTable.updateRadius(enr.nodeId, dataRadius)
+            this.portal.enrCache.updateNodeFromPong(enr, this.networkId, {
+              radius: dataRadius,
+            })
             break
           }
           case PingPongPayloadExtensions.ERROR_RESPONSE: {
```
acolytec3, line 355: I can't comment below but should we do some peer scoring on calls to `routingTable.evictNode`?

ScottyPoi: Yes.  

Currently we don't have any of the peer scoring implemented.  I figured that would come next, after getting this in place.  Not sure which metrics are really worth tracking, or when to actually look at a peer's score, and what decisions to make based on a peer score.  

I do believe that some sort of scoring and peer metrics are helpful, but I think the details of that warrant some discussion.

PR #727, opened 2025-02-03
Network Census

ScottyPoi:
This introdueces a new mapping in the `RoutingTable` class which track the client info from Ping/Pong payloads.  

The network census tracks all known nodes in the network, not just nodes in the kademlia routing table, which has bucket size limits.

A new class called ScoredPeer was created for this mapping.  The currently implementation of ScoredPeer stores minimal information, but the class could be used in future updates to track different metrics or history.

### DIFF
--- a/packages/portalnetwork/src/client/routingTable.ts
+++ b/packages/portalnetwork/src/client/routingTable.ts
@@ -3,16 +3,22 @@ import { KademliaRoutingTable } from '@chainsafe/discv5'
 import type { ENR, NodeId } from '@chainsafe/enr'
 import type { Debugger } from 'debug'
 import { shortId } from '../index.js'
+import { ScoredPeer } from '../networks/peers.js'
 export class PortalNetworkRoutingTable extends KademliaRoutingTable {
   public logger?: Debugger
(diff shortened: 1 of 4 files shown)


packages/portalnetwork/src/client/routingTable.ts
```diff
@@ -2,17 +2,24 @@ import { KademliaRoutingTable } from '@chainsafe/discv5'
...
+import { ScoredPeer } from '../networks/peers.js'
 export class PortalNetworkRoutingTable extends KademliaRoutingTable {
   public logger?: Debugger
   private radiusMap: Map<NodeId, bigint>
   private gossipMap: Map<NodeId, Set<Uint8Array>>
   private ignored: [number, NodeId][]
+  /**
+   * Map of all known peers in the network
+   */
+  public networkCensus: Map<string, ScoredPeer>
```
acolytec3, line 16:
```suggestion
  public networkCensus: Map<NodeId, ScoredPeer>
```

Import `NodeId` from `@chainsafe/enr`

ScottyPoi:
closing to combine with #726

PR #733, opened 2025-02-14
Implement support for ephemeral headers

### LINKED ISSUE #741: Implement support for ephemeral headers
Ephemeral headers are the section of the chain between the current head and end of the post most recent CL sync period. Since the HistoricalSummaries which contains the finalized block roots for each sync period, doesn't have a finalized root for blocks in the current period (since it's not been finalized yet), we keep "ephemeral" headers and provide them in responses to FINDCONTENT messages. Ideally, our set of ephemeral headers should include headers starting with the header contained in the most current optimistic update we know and then contain ancestors of that header back to the first block in the sync period.

Here is all the work I'm aware of related to supporting this: 
To-dos:

- [x] Implement support for storing ephemeral headers in the HistoryNetwork #733
- [x] Maintain an ephemeralHeaderIndex of block numbers to hashes #733
- [x] Report the current number of ephemeral headers contained in our DB in the History_Radius ping/pong payloads #733
  - [x] Add pingPongPayload to HistoryNetwork
  - [x] Remove History_Radius from BaseNetwork
- [ ] Verify that ephemeral headers are only requested for blocks newer than the most recent finalized sync period
- [ ] Add support for returning ephemeral headers in response to FINDCONTENT messages
- [ ]  Purge ephemeral headers once we roll over to a new sync period
- [ ]  Downvote/ban/ignore clients who give invalid responses to ephemeral header FINDCONTENT requests
- [ ] Implement strategy for requesting ephemeral headers from peers when we don't have the whole recent sync period
- [ ] Ensure that we have HistoricalSummaries before requesting ephemeral headers
- [ ] Only accept ephemeral headers that are more recent than latest block in latest finalized sync period
- [ ] Add ephemeral header on each new optimistic update

acolytec3:
Internal work related to support for ephemeral headers - partially fixes #741

- [x] Implement support for storing/retrieving ephemeral headers in the `HistoryNetwork` DB
- [x] Maintain an `ephemeralHeaderIndex` of block numbers to hashes - this is stored in memory for ready access to the headers (which are only stored by blockhash in the DB)
- [x] Report the current number of ephemeral headers contained in our DB in the `History_Radius` ping/pong payloads
  - [x] Add `pingPongPayload` to `HistoryNetwork` - added the History Network specific payload extension here
  - [x] Remove `History_Radius` from `BaseNetwork` - Removed from base network as it doesn't make sense to have specific subnetwork stuff in our `BaseNetwork` class


packages/portalnetwork/src/networks/history/history.ts
```diff
@@ -324,6 +381,68 @@ export class HistoryNetwork extends BaseNetwork {
     }
   }
 
+  protected overridehandleFindContent = async (
```
ScottyPoi, line 384:
```suggestion
  protected override handleFindContent = async (
```

ScottyPoi (approved):
LGTM

PR #736, opened 2025-02-20
Portal Spec Test Runner + SSZ fixes for BlockHeader Proof types

acolytec3:
This PR does 2 things: 
- adds a new test script for running the `portal-spec-test` test vectors, currently hardcoded to run the hive test vectors.  

Since these tests are simple content key -> content value vectors, this can easily be expanded to run any of the spec tests by simply reading in all the `yaml` files from the portal-spec-test package and then attempting to `store` them with the appropriate `network` class.

It currently works for the serialized test vectors in the `history` network though is turned off since it's still incomplete.

- Updates the `History` ssz containers for the block header proofs to match the current spec as we were out of alignment with it.  It also cleans up the associated test data so tests should all pass


ScottyPoi:
We should use this to replace all tests like https://github.com/ethereumjs/ultralight/blob/446a66cae81bd6d287efc71a2cc81b74cb301b51/packages/portalnetwork/test/wire/types.spec.ts where we have the test vectors hardcoded.

packages/portalnetwork/src/networks/history/history.ts
```diff
@@ -219,6 +223,7 @@ export class HistoryNetwork extends BaseNetwork {
       let deserializedProof: ReturnType<typeof HistoricalSummariesBlockProof.deserialize>
       try {
         deserializedProof = HistoricalSummariesBlockProof.deserialize(proof)
+        console.log(HistoricalSummariesBlockProof.toJson(deserializedProof))
```
ScottyPoi, line 226: don't want to do `console.log`

either `this.logger(...)` or delete

packages/portalnetwork/test/networks/history/spec-test-runner.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+import { afterAll, beforeAll, describe, it } from 'vitest'
+import type { HistoryNetwork } from '../../../src/index.js'
+import {
+  HistoryNetworkContentType,
+  PortalNetwork,
+  decodeHistoryNetworkContentKey,
+  getContentKey,
+} from '../../../src/index.js'
+
+describe.skip('should run all spec tests', () => {
```
ScottyPoi, line 15: Maybe insignificant, but I believe that even when tests have `.skip` vitest does the pre-run processing and test set up.  It might be better to instead add this to the `exclude` list in the vitest config flie.

ScottyPoi (approved):
LGMT.  A couple nits left in comments

PR #743, opened 2025-02-28
RPC: Ping extension support

ScottyPoi:
Implements changes in https://github.com/ethereum/portal-network-specs/pull/369/files

- Adds helper method `ultralight_getPingPayload`
  - Params: `NetworkId`, `Extension Type`
  - Returns: Ping extension payload

TODO: 
- Updates `portal_*Ping` with ping extension support

### DIFF
--- a/packages/cli/src/rpc/modules/ultralight.ts
+++ b/packages/cli/src/rpc/modules/ultralight.ts
@@ -1,5 +1,5 @@
 import { hexToBytes } from '@ethereumjs/util'
-import { HistoryNetworkContentType, NetworkId, addRLPSerializedBlock } from 'portalnetwork'
+import { HistoryNetworkContentType, NetworkId } from 'portalnetwork'
 
 import { INTERNAL_ERROR } from '../error-code.js'
 import { middleware, validators } from '../validators.js'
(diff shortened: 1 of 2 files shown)


packages/cli/src/rpc/modules/portal.ts
```diff
@@ -551,52 +559,80 @@ export class portal {
...
-    this.logger(`PING request received on HistoryNetwork for ${shortId(encodedENR.nodeId)}`)
-    const pong = await this._history.sendPing(encodedENR, 0)
+    const extension = ext ?? 0
+    this.logger(
+      `PING request received on HistoryNetwork for ${shortId(encodedENR.nodeId)} with extension ${extension}`,
+    )
+
+    let encodedPayload = undefined
+    if (payload !== undefined) {
+      if (typeof payload === 'string') {
```
ScottyPoi, line 572: is this necessary?  it shouldn't be a possibility i don't think.  and this line is not included in the other network PING methods

acolytec3: No, it shouldn't be.  I was trying to get the RPC tests to work.  Will remove.

acolytec3 (approved):
LGTM

PR #746, opened 2025-03-12
Remove `era`

acolytec3:
Removes the `era` package as it has been superseded by the `@ethereumjs/era` package that will be published shortly.


holgerd77:
What's the status of the `@ethereumjs/era` package, should this be marked as `experimental` or `alpha` (beyond the pre-release versioning)? Or is this sufficiently stable already and ready for production usage, also that we do not substantially (so: in a breaking way) change the API anymore? (so that would then become mandatory, that breaking API changes are only possible/allowed for the next breaking release round)

ScottyPoi:
> What's the status of the `@ethereumjs/era` package, should this be marked as `experimental` or `alpha` (beyond the pre-release versioning)? Or is this sufficiently stable already and ready for production usage, also that we do not substantially (so: in a breaking way) change the API anymore? (so that would then become mandatory, that breaking API changes are only possible/allowed for the next breaking release round)

I think it can be considered stable for production and ready for release

PR #752, opened 2025-03-21
Prune from approximate size

ScottyPoi:
This PR aims to improve and optimize the process of pruning a network DB.  

Instead of measuring / pruning after **random** `OFFER`s, we keep an approximate db measurment updated and prune when the approximate size reaches `maxStorage`

1. Adds `Lock` to prune method
  - boolean `network.pruning` set to `true` during async `prune` method
  -  prune method "locked" while `network.pruning` set to `true`
2. Track `networkDB.approximateSize`
  - `approximateSize` updated during `networkDB.put` and `networkDB.del`
  -  `approximateSize` set to "accurate size" during `networkDB.size()`
3. `network.prune()` triggered when `networkDB.approximateSize > network.maxStorage`
  - compare on `ContentAdded` event
  - no longer prune on random OFFER

----

Also fixed bugs in `prune` method.
- Handle case of non-hex key (like `radius`)
- Handle case of deleting header where block not indexed

### DIFF
--- a/packages/portalnetwork/src/networks/network.ts
+++ b/packages/portalnetwork/src/networks/network.ts
@@ -36,6 +36,7 @@ import {
   ErrorPayload,
   HistoryRadius,
   MAX_UDP_PACKET_SIZE,
+  MEGABYTE,
   MessageCodes,
   NodeLookup,
   PingPongErrorCodes,
@@ -118,7 +119,8 @@ export abstract class BaseNetwork extends EventEmitter {
     }
     this.gossipManager = new GossipManager(this, gossipCount)
     this.on('ContentAdded', () => {
-      if (this.db.approximateSize > this.maxStorage) {
+      if ((this.db.approximateSize / MEGABYTE) > this.maxStorage) {
+        this.logger(`Pruning due to size limit ${this.db.approximateSize / MEGABYTE} > ${this.maxStorage}`)
         void this.prune()
       }
     })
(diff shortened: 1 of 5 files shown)


ScottyPoi:
converting to draft.  i unconvered some bugs in prune that i am working on.

ScottyPoi:
OK fixed bugs in prune and added a test

packages/portalnetwork/src/networks/network.ts
```diff
@@ -178,28 +184,47 @@ export abstract class BaseNetwork extends EventEmitter {
...
       }
-      let size = await this.db.size()
-      const toDelete: [string, string][] = []
+      size = await this.db.size()
       while (size > this.maxStorage * MB) {
+        this.logger.extend('prune')(`Size too large: ${size} > ${this.maxStorage * MB}`)
+        this.logger.extend('prune')(`old radius: ${this.nodeRadius}`)
         const radius = this.nodeRadius / 2n
+        this.logger.extend('prune')(`new radius: ${radius}`)
+        await this.setRadius(radius)
```
acolytec3, line 204: Is there ever a situation where we would want to expand our radius?

acolytec3 (approved):
This looks good.  In general, do we need a mechanism for potentially enlarging our radius?   Could be a future feature where we keep track of how much content we decline based on radius and then expand our radius if we're some percentage under max DB size after like 24 hours or something so we can better serve the network.

PR #755, opened 2025-03-25
client: catch handle TalkReq error

ScottyPoi:
Clients will ERROR and freeze if an incoming TalkReq message contains a payload that cannot be deserialized.

This PR adds a try/catch around `network.handle` so that this sort of error does not kill the client.

### DIFF
--- a/packages/portalnetwork/src/client/client.ts
+++ b/packages/portalnetwork/src/client/client.ts
@@ -424,8 +424,14 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
 
       return
     }
-
-    await network.handle(message, nodeAddress)
+    try {
+      await network.handle(message, nodeAddress)
+    } catch (err: any) {
+      this.logger.extend('error')(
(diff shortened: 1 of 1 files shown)


packages/portalnetwork/src/client/client.ts
```diff
@@ -424,8 +424,14 @@ export class PortalNetwork extends EventEmitter<PortalNetworkEvents> {
...
     }
-
-    await network.handle(message, nodeAddress)
+    try {
+      await network.handle(message, nodeAddress)
+    } catch (err: any) {
+      this.logger.extend('error')(
+        `Error handling TALKREQ message from ${nodeAddress.nodeId}: ${err}.  `,
+      )
+      await this.sendPortalNetworkResponse(nodeAddress, message.id, new Uint8Array())
```
acolytec3, line 433:
```suggestion
      // TODO: Determine if it makes more sense to just not respond versus sending an empty response
      await this.sendPortalNetworkResponse(nodeAddress, message.id, new Uint8Array())
```

acolytec3 (approved):
LGTM

PR #758, opened 2025-03-25
Load ephemeral headers from optimistic updates

acolytec3:


acolytec3:
This PR is no longer valid and being replaced by other work.

PR #763, opened 2025-03-28
Fix RPC logging error

ScottyPoi:
Fix for error in logging of RPC parameters.  `Ping` methods include optional parameters.  We log RPC methods and parameters when called, and this log attempted to do `param.toString()`, which was throwing an error if `param` was null/undefined.

Rather than leave optional parameters empty, the `hive` tests inserts a `None` value.  This is why we ended up trying to do `null.toString()`.  When left empty, the optional parameters are not logged


acolytec3 (approved):
Lgtm

PR #764, opened 2025-03-29
Version 1: Accept-codes / Varint prefix

ScottyPoi:
WIP

Implements first version update utilizing new protocol versioning mechanism.

- Accept Codes
  - Implements accept codes update for OFFER/ACCEPT 
  - v1 ACCEPT message contains `bytelist` instead of `bitlist`
  - `bytelist` usage like previous `bitlist` usage, with each `byte` in the list corresponding to a response to an offered content key
  - Different byte values correspond to different ACCEPT or DECLINE messages
  - Clients can communicate reason for DECLINE

- Versioning
  - versioned types for ACCEPT payload
  - versioned types for SSZ portal wire messages
    - currently only difference between v1 and v2 is `AcceptMessageType`
  - versioned types for `uTP` `ContentRequest`
    - version 1 reads prefix on FindContent 
  - version determines varint prefix for FindContent contents

- FindContent varint prefix
  - version 1 content will have prefix when sent over `uTP`
  - matches version 0 offer/accept content encoding
  - https://github.com/ethereum/portal-network-specs/pull/383/files

### DIFF
--- a/packages/portalnetwork/src/networks/network.ts
+++ b/packages/portalnetwork/src/networks/network.ts
@@ -65,7 +65,7 @@ export abstract class BaseNetwork extends EventEmitter {
     PingPongPayloadExtensions.CLIENT_INFO_RADIUS_AND_CAPABILITIES,
     PingPongPayloadExtensions.BASIC_RADIUS_PAYLOAD,
   ]
-  static MAX_CONCURRENT_UTP_STREAMS = 50
+  public MAX_CONCURRENT_UTP_STREAMS = 50
   public routingTable: PortalNetworkRoutingTable
   public nodeRadius: bigint
   public db: NetworkDB
@@ -765,7 +765,7 @@ export abstract class BaseNetwork extends EventEmitter {
       case 0:
       default: {
         const contentIds: boolean[] = Array(msg.contentKeys.length).fill(false)
-        if (this.portal.uTP.openRequests() > BaseNetwork.MAX_CONCURRENT_UTP_STREAMS) {
+        if (this.portal.uTP.openRequests() > this.MAX_CONCURRENT_UTP_STREAMS) {
           this.logger.extend('OFFER')(`Too many open UTP streams - rejecting offer`)
           return this.sendAccept<0>(src, requestId, contentIds, [])
         }
--- a/packages/portalnetwork/src/networks/network.ts
+++ b/packages/portalnetwork/src/networks/network.ts
@@ -1113,15 +1113,22 @@ export abstract class BaseNetwork extends EventEmitter {
     const offerMsg: OfferMessage = {
       contentKeys: [contentKey],
     }
-    const payload = PortalWireMessageType.serialize({
-      selector: MessageCodes.OFFER,
-      value: offerMsg,
-    })
     const offered = await Promise.allSettled(
       peers.map(async (peer) => {
         this.logger.extend(`gossipContent`)(
           `Offering ${bytesToHex(contentKey)} to ${shortId(peer.nodeId)}`,
         )
+        let version
+        try {
+          version = await this.portal.highestCommonVersion(peer)
+        } catch (e: any) {
+          this.logger.extend('error')(e.message)
+          return
+        }
+        const payload = PortalWireMessageType[version].serialize({
+          selector: MessageCodes.OFFER,
+          value: offerMsg,
+        })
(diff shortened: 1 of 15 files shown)


packages/portalnetwork/src/wire/types.ts
```diff
@@ -134,15 +151,36 @@ export type MessageTypeUnion = [
...
-  PingMessageType,
-  PongMessageType,
-  FindNodesMessageType,
-  NodesMessageType,
-  FindContentMessageType,
-  ContentMessageType,
-  OfferMessageType,
-  AcceptMessageType,
-])
+export const PortalWireMessageType: Record<Version, UnionType<any>> = {
```
acolytec3, line 156: This is really, really nice.

packages/portalnetwork/src/networks/network.ts
```diff
@@ -398,7 +414,7 @@ export abstract class BaseNetwork extends EventEmitter {
     }
   }
 
-  handlePing = async (src: INodeAddress, id: bigint, pingMessage: PingMessage) => {
+  handlePing = async (src: INodeAddress, id: bigint, pingMessage: PingMessage, _version: Version) => {
```
acolytec3, line 417: Why are we adding this here?  Are you just future proofing since we might need it some day?

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/ContentRequest.ts
```diff
@@ -212,14 +229,41 @@ export class FindContentReadRequest extends ContentReadRequest {
...
       // If all packets have been received, return content
-      await this.returnContent([Uint8Array.from(this.socket.reader!.bytes)], [this.contentKey])
+      switch (this.version) {
+        case 0:
+          await this.returnContent([Uint8Array.from(this.socket.reader!.bytes)], [this.contentKey])
+          break
+        case 1: {
+          if (this.socket.reader!.contents.length > 0) {
+            const key = this.contentKey
+            const value = this.socket.reader!.contents.shift()!
```
acolytec3, line 239: Why is this happening?  Are we assuming there's no more content to be received?

packages/portalnetwork/src/wire/utp/Socket/ReadSocket.ts
```diff
@@ -42,10 +42,7 @@ export class ReadSocket extends UtpSocket {
       expected = this.ackNr + 1 === packet.header.seqNr
     }
     this.setSeqNr(this.getSeqNr() + 1)
-    if (!this.reader) {
-      this.reader = new ContentReader(packet.header.seqNr)
-      this.reader.bytesExpected = Infinity
-    }
+
```
acolytec3, line 45: Are we confident there's never a place where a reader won't have been instantiated?

ScottyPoi: I moved this line into the request class, since it is aware of the version and the socket is not

packages/portalnetwork/src/wire/utp/PortalNetworkUtp/ContentRequest.ts
ScottyPoi: Yes, this is a FindContent request, so 1 piece of content.  since we now know the length of that content via the prefix, we know when we have received the whole stream even before receiving the FIN

packages/portalnetwork/src/networks/network.ts
```diff
@@ -240,30 +241,38 @@ export abstract class BaseNetwork extends EventEmitter {
   public async handle(message: ITalkReqMessage, src: INodeAddress) {
     const id = message.id
     const request = message.request
-    const deserialized = PortalWireMessageType.deserialize(request)
+    const enr = this.findEnr(src.nodeId)!
+    let version: Version
+    try {
+      version = await this.portal.highestCommonVersion(enr)
+    } catch (e: any) {
```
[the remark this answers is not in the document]
ScottyPoi: No i think you're right that they are redundant

packages/portalnetwork/src/networks/network.ts
ScottyPoi: Yes.  we could remove it if that is confusing

acolytec3: It's fine. Just wasn't sure what it was doing there

PR #765, opened 2025-04-02
PortalClient: PoC for PortalClient

cjustinobi:
This PR introduces cross platform interface for interacting with Ultralight Portal Network.

#### What this PR does

 - User should be able to initialize PortalNetwork node
 - User can query `eth_getBlockByNumber` and `eth_getBlockByHash` via nice interface on mobile device and desktop
 - Set up the test suite with Vitest and with the necessary test libraries
 - Unit tests covering implemented RPC methods
 - Intuitive alert component for error handling and notifications
 
 PS: The commits could be squashed from the merge point if everything looks good👌


packages/portal-client/README.md
ScottyPoi, line 1: Please add startup instructions here

packages/portalnetwork/src/client/constructor.ts
```diff
@@ -100,7 +100,7 @@ export async function createPortalNetwork(opts: Partial<PortalNetworkOpts>): Pro
       break
     }
     case TransportLayer.MOBILE:
```
ScottyPoi, line 102:
```suggestion
    case TransportLayer.TAURI:
```
Let's add `TAURI` to the `TransportLayer` enum and use this case instead of `MOBILE`

packages/portalnetwork/src/util/config.ts
```diff
@@ -36,6 +36,25 @@ export const NetworkStrings: Record<string, NetworkId> = {
   state: NetworkId.StateNetwork,
 }
 
+let execSync: Function
```
ScottyPoi, line 39: What is the purpose of all of this?  Do we really need this?

test.sh
ScottyPoi, line 1: What is the purpose of this file?
It seems to be specific to "Justin's Macbook Pro"

ScottyPoi (commented):
Nice work, this is awesome!  There are a few things I don't quite understand, and some nits left in comments, but this is coming along great.

test.sh
cjustinobi: My bad. Specific to me. Will update

holgerd77:
Is this thing now actually working, so in "real life"?

Can we get a screenshot or something? 🙂 

And some at least minimal doc section for the README would also be helpful/necessary (or did I overlook?).

Beside: I am really not sure if this should really stay inside the Ultralight monorepo, added quite a lot of code/files/whatever. Seems rather a candidate for a separate repo (?). Somewhat additional topic though.

acolytec3:
> Is this thing now actually working, so in "real life"?
> 
> Can we get a screenshot or something? 🙂
> 
> And some at least minimal doc section for the README would also be helpful/necessary (or did I overlook?).
> 
> Beside: I am really not sure if this should really stay inside the Ultralight monorepo, added quite a lot of code/files/whatever. Seems rather a candidate for a separate repo (?). Somewhat additional topic though.

It **was** working last time I looked at it.  It's having trouble connecting to bootnodes when I run it locally so will need to do some additional research as to why that is. 

It's still pretty basic at the moment, basically just a GUI wrapper for the RPC but I think the promise is there.

holgerd77:
OK but even just GUI wrapper it would be nice to have working (and baseline-documented to be used (e.g. by us 🙂, team - (Scotty + Andrew)))

PR #767, opened 2025-04-09
PortalClient: Rebuild node from memory

cjustinobi:
This PR improves the application by persisting the node details
to prevent fresh node creation on every initialisation.
 - Added dedicated methods for private key management (`getPrivateKey`, `savePrivateKey`)
 - Moved a couple of code to the proper files.


acolytec3 (approved):
LGTM.

Note, I added a couple of commits.
1) Remove the debug shim as this is unnecessary.
2) Added the `portalnetwork` client as an object on the `Window` as `Window.portal` so you can access the portalnetwork client directly for debugging.

acolytec3 (approved):

Issue #711, opened 2025-01-16
Ensure `contentLookup` never has more than 5 active requests


acolytec3:
I've noticed in the course of testing stuff on the live network that our `contentLookup` is queueing up more than 5 requests and gets stuck in this pending state where there are up to 20 open requests and none of them have resolved.

One example.  Easily reproducible if you request the `LightClientBootstrap` on the beacon network with a local node.

![Image](https://github.com/user-attachments/assets/db4f5023-019e-4349-ae55-56765c8b00e8)

We should try and reproduce this in a test locally and then find a fix.

The likely solution involves adding a timeout around in [this section of code](https://github.com/ethereumjs/ultralight/blob/c5446855976fe36f528b57bebfd81c0d484a6589/packages/portalnetwork/src/networks/contentLookup.ts#L158) that will delete the request if the other node doesn't respond within a certain amount of time.

acolytec3:
2 things to reproduce
1) To see the `ContentLookup` logs, you need to add it to the debug topics.  If you run it with `DEBUG=*BeaconNetwork:*,ultralight*,*ContentLookup* npm run dev-testnet -- --rpc`, you'll be able to see content lookup logs.
2) To reproduce the actual issue, you don't actually need to sync the beacon network, just use the `portal_beaconGetContent` method to do a recursive content lookup.  This command `curlie 127.0.0.1:8545 jsonrpc=2.0 id=1 method=portal_beaconGetContent params:='["0x10f090380ffd4ba91c51488bb0fcdcfe70a34d03fac8011861f65bd9ef90576639"]'` produced the issue for me.

![Image](https://github.com/user-attachments/assets/d60d9f16-bb2d-4277-89ae-17b92399d835)

A proper bug fix will make sure that open requests are cleared out of the queue once they resolve or timeout.

foufrix:
Thanks I was able to reproduce as well

![Image](https://github.com/user-attachments/assets/69a90431-082b-4b72-9484-84a14f87f3b9)

Sometimes it effectively retry periodically with error and the curlie call is on hold, but sometime it returns `Internal Error`

![Image](https://github.com/user-attachments/assets/189636b2-0446-41bd-97f8-f000da9d05c9)

I suppose the goal here is to prevent retries when it does not work, always return internal error when there is an issue and avoid having multiple RPC failure? 

![Image](https://github.com/user-attachments/assets/6c5e1f0d-9ce6-464e-a367-35f55c7ea811)

[...]

PR #780, opened 2025-05-05
ContentLookup bug fix

ScottyPoi:
Fix for ContentLookup bug described in #711 

Concurrent requets should be limited to 5, however, we were often seeing logs like: 
```
7e7fd:Portal:BeaconNetwork:LOOKUP:0x10f090… Have 19 peers left to ask and 10 pending requests
```

This happened because any time a peer request returned (without content), it would trigger the next recursive round of requets, which would add 5 new requests to the batch.

However, we do want to avoid the need to wait for full batches of 5 to complete before continuing the recursive search.

The solution is to check the number of pending requests before adding more to the batch, and to limit the total number of requests in the new batch and old to 5.  We accomplish this with this line:
```
        const peerBatch: LookupPeer[] = []
        const availableSlots = 5 - this.pending.size
        while (this.lookupPeers.peek() && peerBatch.length < availableSlots) {
          const next = this.lookupPeers.pop()!
          peerBatch.push(next)
        }
```

----

This solution did limit the total pending requests to 5, but did not solve another underlying issue of incomplete requests hanging open.  Because of this, we will simply end up with a set of 5 pending requests that never close.

To solve this, we couple each request with an `abortPromise` using [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController), and `AbortSignal`.  The promise will reject when the signal is aborted.  The actual content request is raced against this abort promise.  If the signal is aborted, the request is cancelled and cleaned up, and the pending request is removed from the tracking set.  

This seems to solve the hanging request issue, which should also solve some memory leaks and resource exhaustion issues.

### DIFF
--- a/packages/portalnetwork/src/networks/contentLookup.ts
+++ b/packages/portalnetwork/src/networks/contentLookup.ts
@@ -92,8 +92,15 @@ export class ContentLookup {
           peerBatch.push(next)
         }
         const promises = peerBatch.map((peer) => {
+          const controller = new AbortController()
+          const timeoutId = setTimeout(() => {
+            controller.abort()
+          }, this.timeout)
+          
           return Promise.race([
-            this.processPeer(peer),
+            this.processPeer(peer, controller.signal).finally(() => {
+              clearTimeout(timeoutId)
+            }),
             new Promise((resolve) =>
               setTimeout(() => {
                 resolve(undefined)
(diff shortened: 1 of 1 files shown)


packages/portalnetwork/src/networks/contentLookup.ts
```diff
@@ -25,6 +25,7 @@ export class ContentLookup {
   private finished: boolean
   private content: ContentLookupResponse
   private pending: Set<NodeId>
+  private queuedPeers: Set<NodeId>
```
foufrix, line 28: How do you remove `queuedPeers`? It seems like it will grow forever

ScottyPoi: `queuedPeers` should always be a full list of all the peers ever added to the `lookupPeers` heap, so that we can avoid re-adding a peer to the heap after it has been popped

acolytec3 (approved):
Lookups are no longer running for ever and queues growing infinitely so LGTM!

packages/portalnetwork/src/networks/contentLookup.ts
foufrix: ok thanks for the clarification! great fix

PR #785, opened 2025-05-13
Support for test chains

ScottyPoi:
This PR adds support for test chains, such as Sepolia or AngelFood

- Adds `ChainId` as client attribute
- Adds chain id to CLI arguments
- Implements `NetworkId` resolution by `ChainId`
- Refactors network constructors
- Refactors client constructor
- Adds chain specific configuration in RPC modules


packages/cli/src/rpc/modules/eth.ts
```diff
@@ -26,7 +26,7 @@ export class eth {
    */
   constructor(client: PortalNetwork, logger: Debugger) {
     this._client = client
-    this._history = client.networks.get(NetworkId.HistoryNetwork) as HistoryNetwork
```
acolytec3, line 29: Want to add `eth_chainId` to this since we're now using `chainId`?  This would also fix a pain point trying to integrate with ponder.

acolytec3 (approved):
LGTM!

PR #787, opened 2025-05-14
Electra Historical Summaries

ScottyPoi:


acolytec3 (commented):
Unfortunately, the failing test isn't going to be fixable without updating the test data.  
Lemme see how hard it is to just use the stuff from the spec test repo.

acolytec3 (approved):

Issue #729, opened 2025-02-05
Implement `trace_offer`


acolytec3:
We should implement a `portal_*TraceOffer` endpoint so we can participate in some benchmarking that's been developed by @KolbyML.

Behavior should follow the existing `portal_*Offer` endpoint except return a trace object that looks like this (in Rust) 
```
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub enum OfferTrace {
    /// Offer was successful, all accepted content keys in bitlist were transferred
    Success(BitList<typenum::U64>),
    /// Peer is not interested in any of the offered content keys
    Declined,
    /// This offer failed, perhaps locally or from a timeout or transfer failure
    Failed,
}
```

PR #795, opened 2025-10-28
feat: add trace-offer

foufrix:
Implement traceOffer function.

It should return an object with this form: 
```ts
export type OfferTrace = {
  success?: boolean[]
  declined?: boolean
  failed?: boolean
}
```

Function have exactly the same behavior as offer, the difference is on what it returns.

A helper function to convert bit to boolean array has been added to improve reusability. For the moment it's in `portal.ts` maybe there is a better place to put this helper?

Test case has been added where we try 3 cases:
- success
- peer not interested
- peer not reacheable (wrong enr)

This solve #729


packages/cli/src/rpc/modules/portal.ts
```diff
@@ -29,6 +29,36 @@ import type { Debugger } from 'debug'
 import type { BeaconNetwork, HistoryNetwork, PortalNetwork, StateNetwork } from 'portalnetwork'
 import type { GetEnrResult } from '../schema/types.js'
 
+/**
```
ScottyPoi: can we move this to `rpc/utils.ts`?

packages/cli/test/rpc/portal/historyTraceOffer.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+      [contentKey1, content1],
+      [contentKey2, content2],
+    ]
+
+    try {
+      const res = await rp.request(method, [enr, contentItems])
+
+      // Should return success with boolean array indicating which keys were accepted
+      assert.ok(res.result.success !== undefined, 'should have success array')
+      if (res.result.success === true) {
```
ScottyPoi: `res.result.success` is a boolean array, so this part gets skipped

you've already asserted that it is not `undefined`, so you can just remove this conditional and let the other checks happen

packages/cli/src/rpc/modules/portal.ts
```diff
@@ -1223,6 +1270,79 @@ export class portal {
...
+    try {
+      if (this._history.routingTable.getWithPending(enr.nodeId)?.value === undefined) {
+        const res = await this._history.sendPing(enr)
+        if (res === undefined) {
+          return { failed: true }
+        }
+      }
+
+      const res = await this._history.sendOffer(enr, contentKeys, contentValues)
+      return bitToBooleanArray(res, contentKeys.length)
```
ScottyPoi: The response will only be a BitArray if running protocol V0

With V1 this type was changed to Uint8Array.  The current state of Ultralight is still designed to support both.  

But if the response is V1 you won't want to call this helper function, which would actually end up returning a full array of accepts.

packages/cli/src/rpc/modules/portal.ts
```diff
@@ -1223,6 +1270,79 @@ export class portal {
...
+    try {
+      if (this._state.routingTable.getWithPending(enr.nodeId)?.value === undefined) {
+        const res = await this._state.sendPing(enr)
+        if (res === undefined) {
+          return { failed: true }
+        }
+      }
+
+      const res = await this._state.sendOffer(enr, contentKeys, contentValues)
+      return bitToBooleanArray(res, contentKeys.length)
```
ScottyPoi: Same here as history method

packages/cli/src/rpc/modules/portal.ts
```diff
@@ -1223,6 +1270,79 @@ export class portal {
...
+    try {
+      if (this._beacon.routingTable.getWithPending(enr.nodeId)?.value === undefined) {
+        const res = await this._beacon.sendPing(enr)
+        if (res === undefined) {
+          return { failed: true }
+        }
+      }
+
+      const res = await this._beacon.sendOffer(enr, contentKeys, contentValues)
+      return bitToBooleanArray(res, contentKeys.length)
```
ScottyPoi: same here as history method

packages/cli/src/rpc/modules/portal.ts
```diff
@@ -29,6 +29,36 @@ import type { Debugger } from 'debug'
 import type { BeaconNetwork, HistoryNetwork, PortalNetwork, StateNetwork } from 'portalnetwork'
 import type { GetEnrResult } from '../schema/types.js'
 
+/**
+ * Converts a BitArray response to a boolean array indicating which content keys were accepted
+ * @param res The response from sendOffer (can be BitArray, undefined, or empty array)
+ * @param contentKeysLength The number of content keys that were offered
+ * @returns Object with success array, declined flag, or failed flag
+ */
+function bitToBooleanArray(res: any, contentKeysLength: number): { success?: boolean[]; declined?: boolean; failed?: boolean } {
```
ScottyPoi:
```suggestion
function bitToBooleanArray(res: BitArray, contentKeysLength: number): { success?: boolean[]; declined?: boolean; failed?: boolean } {
```

ScottyPoi (commented):
Thanks for the PR!
We have not been developing on Ultralight recently, as we evaluate the future of Portal Network in general.  Reviewing this with some optimism that some version of Portal will still be used by EL clients.

Good work overall.  I left a few comments to address.  There's a bug in how the new RPC methods are dealing with the response.  I'll take some blame there, as the upgrade to protocol V1 was not well documented in the code.  The upgrade from v0 to v1 involved changing the return type of the contentKeys array in an ACCEPT response.  With v0 we used a BitArray, and with V1 it is a Unit8Array.  Currently things are set up to support both versions, so we'll need to handle the reponse differently if it's a BitArray or Uint8Array for now, and you won't want to call your helper function on a Uint8Array response.

packages/cli/src/rpc/modules/portal.ts
foufrix: My bad I may have done that during test because I had issue and forgot to put it back thanks 🙏

packages/cli/src/rpc/modules/portal.ts
foufrix: ok, let me see and update accordingly to support both

packages/cli/src/rpc/modules/portal.ts
foufrix: thanks that's the info I was looking for 🙏

packages/cli/src/rpc/modules/portal.ts
```diff
@@ -1223,6 +1241,106 @@ export class portal {
...
+      if (res === undefined || res === null || Array.isArray(res) && res.length === 0) {
+        return { declined: true }
+      }
+      if (res instanceof BitArray) {
+        return bitToBooleanArray(res, contentKeys.length)
+      }
+      if (res instanceof Uint8Array) {
+        return Uint8toBooleanArray(res, contentKeys.length)
+      }
+      return { failed: true }
```
foufrix, line 1271: @ScottyPoi should I return here failed or a full length success
```js
return { success: new Array(contentKeysLength).fill(true) }
```

foufrix:
@ScottyPoi thanks for the review:
- test case fixed by removing useless if statement
- function move to rpc/utils
- new function to manage UInt8 to boolean array
- trigger the correct function depending of `typeof res`

Can you confirm what should be return in case we don't have `UInt8Array` or `BitArray` 🙏 

Also I should not continue to take other issue if project is on halt? Where can I follow the state of decision for the future of Portal Network?

packages/cli/src/rpc/util.ts
```diff
@@ -22,3 +23,39 @@ export function callWithStackTrace(handler: Function, debug: boolean) {
...
+  for (let i = 0; i < contentKeysLength; i++) {
+    const byteIndex = Math.floor(i / 8)
+    const bitIndex = i % 8
+    const byte = res[byteIndex]
+    if (byte !== undefined) {
+      successArray[i] = ((byte >> bitIndex) & 1) === 1
+    }
+  }
+  return { success: successArray }
+}
```
ScottyPoi, line 61: I think this is giving false results.  With the V1 response, each byte in the Uint8Array represents an `AcceptCode`.  `AcceptCode.ACCEPT = 0`, and all other values represent different types of rejections.  So for each index of the Uint8Array, just check if `res[i] === AcceptCode.ACCEPT`, and that should give you the boolean array.
```suggestion
export function Uint8toBooleanArray(res: Uint8Array, contentKeysLength: number): { success: boolean[] } {
  const successArray = new Array(contentKeysLength).fill(false)
  for (let i = 0; i < contentKeysLength; i++) {
    if (byte !== undefined) {
      successArray[i] = AcceptCode.ACCEPT
    }
  }
  return { success: successArray }
}
```

packages/cli/test/rpc/portal/historyTraceOffer.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    // Generate content keys that are close to node2's ID to increase acceptance probability
+    const contentKey1 = '0x00' + '5'.repeat(64)
+    const content1 = '0x' + '1'.repeat(128)
+    const contentKey2 = '0x00' + '6'.repeat(64)
+    const content2 = '0x' + '2'.repeat(128)
+
+    // Store content in node2
+    await rp2.request('portal_historyStore', [contentKey1, content1])
+    await rp2.request('portal_historyStore', [contentKey2, content2])
```
ScottyPoi, line 68: If node2 stores these contents first, it will reject them in the OFFER.  This test is giving you a false positive because of the bug in the `Uint8ToBooleanArray` helper

ScottyPoi:
> Can you confirm what should be return in case we don't have `UInt8Array` or `BitArray` 🙏

`sendOffer` will return `undefined` if there is a version mismatch, or throw an error if the message doesn't send for some reason.  I think we would return "failed" in these circumstances, but double check the specs for this method.  


> Also I should not continue to take other issue if project is on halt? Where can I follow the state of decision for the future of Portal Network?

The original set of Portal Networks are now "Legacy" and no longer live.  The only Portal Network currently in development is a new, minimal version of history.  I'm going to rename all of the old networks with a `Legacy` tag.  I don't want to remove them, in case there is a revival at some point.

The work you're doing now is still helpful though!  I plan to implement the new `ExecutionHistoryNetwork`, which you're welcome to help with, and adding a `traceOffer` method for that network will be simple since

packages/cli/src/rpc/modules/portal.ts
ScottyPoi: I think "failed" here -- since the only other response would be `undefined` if we can't actually send this peer an Offer due to version mismatch

foufrix: Thanks it confirms what I was thinking

